PR #36, opened 2020-10-23
make light IDE UI part of the build

akphi:
@pierredebelen In this PR, I setup the build for `legend-pure-ide-light` so that it also do NPM build for the IDE UI. Also, I reconfigure the UI code to read the config from `ideLightConfig.json`. With this setup, the user only needs to build the module and start Dropwizard server, then they can access the editor at `http://localhost:9010/ide` 🍰

PR #38, opened 2020-10-26
Upgrade dropwizard and jetty

kevin-m-knight-gs:


akphi:
@kevin-m-knight-gs I didn't realize we already have this, I included this in #36 to make pass whitesource check, can you close this one?

PR #47, opened 2020-10-29
added getAllForEachDate function

gs-gunjan:


gs-gunjan:
Found the bug, will create another request

PR #52, opened 2020-11-04
fix java code gen bug

hardikmaheshwari:
Identified a bug in java code gen flow. It effects graph queries working on mapping having sets with and without explosion property mapping


hardikmaheshwari:
@cla-bot[bot] check

PR #56, opened 2020-11-25
Pure IDE: add config for source file locations

aziemchawdhary-gs:
Make locations of Pure core source file configurable and allow option of keeping welcome.pure in persistent location in file system.
New configuration options are completely optional and not providing these configuration options will fallback to existing behaviour.


pierredebelen (approved):

pierredebelen:
Will fix the CVE (which is not related) in a different commit

PR #57, opened 2020-11-25
Check function expressions in constraints

aziemchawdhary-gs:
Fixes the Pure compiler to type check lambdas in constraint definitions (both function and message lambdas).


aziemchawdhary-gs:
This change will allow us to remove the Logging warning in [legend-engine](https://github.com/aziemchawdhary-gs/legend-engine/blob/ca3f4538c22b99feda8704746a74799ff1fb9b0e/legend-engine-language-pure-compiler/src/main/java/org/finos/legend/engine/language/pure/compiler/toPureGraph/HelperModelBuilder.java#L211), and return an error message instead of allowing invalid constraints to pass compiler checks. Thus Studio will be able to warn users that they have defined an invalid constraint. 

This will likely affect existing models which currently have constraints which are not correct with respect to types and will require some work to fix those before this change is merged.

pierredebelen (approved):

pierredebelen:
Will fix the CVE (which is not related) in a different commit

PR #65, opened 2020-12-07
Add two core String functions.  isLetter() and isAlphaNumeric() 

YannanGao-gs:
Add two core String functions. isLetter() is to check if a string only consists of letters, isAlphaNumeric() is to check if string is alphaNumeric.


akphi:
@cla-bot[bot] check

YannanGao-gs:
@cla-bot[bot] check

PR #99, opened 2021-01-26
Support a new PURE primitive type StrictTime

YannanGao-gs:
Support a new PURE primitive type StrictTime


YannanGao-gs:
An exiting test needs a fix since m3.pure has been changed.

PR #291, opened 2021-08-10
fix Model Chain Connection use case with result sourcing

hardikmaheshwari:
fix Model Chain Connection use case with result sourcing


pierredebelen:
I'm not sure to understand the test. How do the two lines below correlate?
```
let x = _Firm.all()->toOne()->graphFetch(#{_Firm {name} }#);
Target_Firm.all()->graphFetch($tree)->serialize($tree);};
```

PR #298, opened 2021-08-16
Java M2M: Add support for calling new from qualified properties

aziemchawdhary-gs:
Expand support for where new is supported: allow use of new in qualified properties provided the return type of the qualified property is primitive. This prevents the use of new to create instances as the result of mapping but allows the creation of ephemeral data for use in validations during mapping execution.


dave-wathen (approved):
Changes look good to me

PR #299, opened 2021-08-17
Java M2M: enable Jackson BigDecimal option if model contains Decimal …

aziemchawdhary-gs:
Enable Jackson option to use BigDecimal for float values in M2M code generation.
Tests will be added to legend-engine as Pure JSON deserializer incorrectly deserializes large BigDecimal values

### DIFF
--- a/legend-pure-code-compiled-core/src/main/resources/core/pure/executionPlan/javaPlatform/executionNode/graphFetch/inMemory/graphFetchJson.pure
+++ b/legend-pure-code-compiled-core/src/main/resources/core/pure/executionPlan/javaPlatform/executionNode/graphFetch/inMemory/graphFetchJson.pure
@@ -43,6 +43,11 @@ function meta::alloy::runtime::java::graphFetch::json::createJsonReading(pureCla
       ->addMavenDependency('com.fasterxml.jackson.core', 'jackson-core', '2.10.3');
 }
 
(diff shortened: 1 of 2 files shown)


dave-wathen (approved):
Changes look good to me

pierredebelen (changes requested):
To avoid confusion, could we translate the Pure Decimal type (fixed point) to BigDecimal?

aziemchawdhary-gs:
> To avoid confusion, could we translate the Pure Decimal type (fixed point) to BigDecimal?

We do use BigDecimal for PURE's Decimal.  This change ensures Jackson internally parses properly when we use them.

pierredebelen:
So what does useBigDecimalForFloats mean?

aziemchawdhary-gs:
sorry - where do you see that? I can't see that in my change - I did have that when I was prototyping the fix, so I may have inadvertently added that.

aziemchawdhary-gs:
> So what does useBigDecimalForFloats mean?

ah I see it - that parameter could be named better I guess

aziemchawdhary-gs:
That flag is used to set the parsing option in Jackson - in Jackson the flag is called USE_BIG_DECIMAL_FOR_FLOATS

pierredebelen (approved):

PR #300, opened 2021-08-19
Graphfetch: validate property owners when computing sourceTree

aziemchawdhary-gs:
When computing properties to include in the source tree we need to make sure that we only include properties that are part of the current class being processed. 
This fix adds validation and filters to ensure that we only include the relevant properties based on the current class being processed.


dave-wathen (approved):

aziemchawdhary-gs:
@pierredebelen this change should have been part of https://github.com/finos/legend-pure/pull/298

PR #294, opened 2021-08-13
Fixed a bug in sourceTreeCalculation when subType is used in m2m mapping

YannanGao-gs:

PR #302, opened 2021-08-24
Revert "Fixed a bug in sourceTreeCalculation when subType is used in …

aziemchawdhary-gs:
…m2m mapping (#294)"

This reverts commit 5d5a81f2ff8287674d480fd00150bfe46ba7ff28.


aziemchawdhary-gs:
Revert required as this breaks existing models and we can no longer generate execution plans for M2M in certains cases.
Will require investigation to work out root cause and fix up the original commit

PR #308, opened 2021-08-31
Add subtype class discoverability using classname for JSON deserialization

rafaelbey:


rafaelbey:
@dave-wathen @AFine-gs -

PR #323, opened 2021-09-17
handle `ClassNotFoundException` in `subTypeOf()`

akphi:
### Context

`ClassNotFoundException` can occur when we use `subTypeOf()` in `engine` where some Java classes are not available during plan generation. There is a _potentially_ less performant alternative which is to use `type_subTypeOf()` as this will use the metamodel graph instead of Java classes to test subtype; but this alternative is more reliable. As such, to be defensive, we should fallback to the latter when the former fails with `ClassNotFoundException`

Full context could be found in https://github.com/finos/legend-pure/issues/324

---

### Other Consideration
- [ ] ~~Do we need to take care of any other cases where the method `pureTypeToJavaClass` is being used?~~
- [ ] Bring the test case in https://github.com/finos/legend-pure/issues/324 to `finos/legend-engine` - check with @aziemchawdhary-gs about this

---

Closes https://github.com/finos/legend-pure/issues/324

- [x] Also took care of `instanceOf()` check
 
@kevin-m-knight-gs Could you kindly take a look at this?

### DIFF
--- a/legend-pure-code-compiled-core/src/main/resources/core/pure/graphFetch/graphExtension.pure
+++ b/legend-pure-code-compiled-core/src/main/resources/core/pure/graphFetch/graphExtension.pure
@@ -117,7 +117,7 @@ function <<access.private>> meta::pure::graphFetch::xstoreProperties(nodes:Graph
       ->cast(@meta::pure::mapping::xStore::XStorePropertyMapping)
       ->map({xpm |
          let allProperties = $xpm.crossExpression.expressionSequence->evaluateAndDeactivate()->toOne()->findPropertiesInValueSpecification();
-         let srcProperties = $allProperties->filter(x | $x.owner->cast(@Class<Any>)->subTypeOf($srcClass) || $srcClass->subTypeOf($x.owner->cast(@Class<Any>)));
+         let srcProperties = $allProperties->filter(x | $x.owner->cast(@Class<Any>)->_subTypeOf($srcClass) || $srcClass->_subTypeOf($x.owner->cast(@Class<Any>)));
          $srcProperties->filter(x | !$currentSubTreeProperties->contains($x));
       });
 }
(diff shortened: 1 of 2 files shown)


kevin-m-knight-gs:
Another place we might want to have such logic is in Pure.instanceOf. We would run into the same problem there under the same circumstances.

legend-pure-code-compiled-core/src/main/resources/core/pure/graphFetch/graphExtension.pure
```diff
@@ -487,7 +487,7 @@ function <<access.private>> meta::pure::graphFetch::propertyTreeToGraphFetchTree
        let childrenGroupedByDerivedClasses = $pTree.children->groupBy(c|$c.value->cast(@PropertyPathNode).class);
        $childrenGroupedByDerivedClasses->keyValues()->map(k|let propertyPathOwner = $k.first;
                                                             let children = $k.second.values;
-                                                            if($propertyPathOwner != $nodeClass && $propertyPathOwner->_subTypeOf($nodeClass->toOne()),
+                                                            if($propertyPathOwner != $nodeClass && $propertyPathOwner->subTypeOf($nodeClass->toOne()),
                                                                 |^PropertyGraphFetchTree(property = $node.property, subType = $propertyPathOwner, subTrees = $children->map(c| $c->propertyTreeToGraphFetchTree($propertyPathOwner))),
```
aziemchawdhary-gs, line 491: We should probably change other uses of _subTypeOf to subTypeOf in this file to be consistent.

akphi: updated!

pierredebelen:
Not sure we should move to subTypeOf if we know we won't find the type reflectively. Doesn't it mean we would pay for the cost of exception management each time?

aziemchawdhary-gs:
> Not sure we should move to subTypeOf if we know we won't find the type reflectively. Doesn't it mean we would pay for the cost of exception management each time?

This is a good point. 

I guess the intention of this change should be to modify subTypeOf is to handle the case when someone inadvertently uses subTypeOf instead of _subTypeOf and we won't fail at runtime if they do. But we should continue to use _subTypeOf
Is that right?

akphi:
@aziemchawdhary-gs I have reverted my changes in the `pure` file, as @pierredebelen has pointed out, using `subTypeOf()` there means that we will bump into the exception and the fall back all the time (which is less performant)

PR #338, opened 2021-09-27
Resolve the imports inside function: Keep the full path of non-meta Applied Function

YannanGao-gs:
Resolve the imports inside function: Keep the full path of non-meta Applied Function


YannanGao-gs:
Code review was approved by Adam Fine.

PR #363, opened 2021-10-13
Parameter external source metamodel

gs-ssh16:


gs-ssh16:
This PR adds support for having function/service parameters as sources of external format connections

PR #379, opened 2021-11-03
Incorrect Snowflake Query function corresponding to jsonStrings functions in Pure

mrudula-gs:
Incorrect Snowflake Query function string_agg corresponding to jsonStrings function in Pure. It should be listagg


davidharte-gs (changes requested):
Please confirm why we are using two different Snowflake functions for joinStrings, in the mapping vs the PureFunction, also the PureFunction tests should access model properties. Thx

PR #381, opened 2021-11-04
Fix and add test for BigQuery dyna concat translation.

CptTeddy:
- Fix and add test for BigQuery dyna concat translation.


Yasirmod17 (approved):

davidharte-gs (commented):
Approved

PR #393, opened 2021-11-22
Reverse milestoning transformation on the property when transforming mapping

YannanGao-gs:
Reverse milestoning transformation on the property when transforming mapping


YannanGao-gs:
Draft this pr as it is the same as https://github.com/finos/legend-pure/pull/394

PR #394, opened 2021-11-23
revertMilestoningTransform on properties when serializing mapping to protocol

vikask-gs:
…protocol


davidharte-gs (changes requested):
As discussed with Yannan we need a test for this change and/or a discussion on why there is none

davidharte-gs (approved):
Approved, please follow through on the follow ups, on a separate PR.

PR #400, opened 2021-12-03
Fix "Not In" Query Generation.

CptTeddy:
Add catch-all fix for "Not In" query generation for MemSQL, DB2, Sybase IQ, Snowflake DB types (RelationalBlockExecutionNode), symmetric to how "In" is handled.


CptTeddy:
Note: tested sql execution in Snowflake console with success.

davidharte-gs (approved):
Approved

PR #404, opened 2021-12-09
Support DROP clause for MemSQL using ROW_NUMBER

mrudula-gs:
Adding support to DROP clause for MemSQL DB type using the window function ROW_NUMBER


davidharte-gs (commented):
Approved

PR #405, opened 2021-12-13
Update previous vx_x_x versioned alloy test to latest pure protocol version.

CptTeddy:
Update previous vx_x_x versioned alloy test to latest pure protocol version.


hardikmaheshwari (approved):

davidharte-gs (approved):
Approved

PR #421, opened 2022-01-17
Initial version of persistence model.

davidsunglee:
Specification for writing the result of an engine service call to a persistent store.

### DIFF
--- a/legend-pure-code-compiled-core/src/main/resources/core/pure/persistence/metamodel.pure
+++ b/legend-pure-code-compiled-core/src/main/resources/core/pure/persistence/metamodel.pure
@@ -1,6 +1,7 @@
 import meta::legend::service::metamodel::*;
 
 import meta::pure::persist::metamodel::*;
+import meta::pure::persist::metamodel::backend::*;
 import meta::pure::persist::metamodel::batch::*;
 import meta::pure::persist::metamodel::batch::deduplication::*;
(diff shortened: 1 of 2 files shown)


legend-pure-code-compiled-core/src/main/resources/core/pure/persistence/metamodel.pure
```diff
@@ -0,0 +1,409 @@
...
+
+/**********
+ * batch mode: milestoned - delta
+ **********/
+
+Class
+<<typemodifiers.abstract>>
+{doc.doc = 'Values in the incoming batch represent a delta to be applied on top of the existing dataset non-destructively. New incoming values represent a logical create; updated incoming values represent a logical update; missing incoming values represent a no-op in the transaction (processing) dimension. Additional merge semantics, including logical deletes, can be supported via a merge scheme.' }
+meta::pure::persist::metamodel::batch::milestoned::delta::DeltaMilestoned extends Milestoned
+{
```
rafaelbey, line 376: Is this missing the field to define the MergeScheme?  Otherwise, the model does not seem to leverage MergeScheme classes.

davidsunglee: Good catch -- it's added in the latest commit.

rafaelbey (approved):
Looks good to me!

PR #430, opened 2022-01-26
Adding nexus staging plugin

ivan-kyosev-gs:
Adding the nexus staging maven plugin to allow for easy deployments to Maven Central

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -30,6 +30,8 @@
     <packaging>pom</packaging>
 
     <modules>
+        <module>legend-pure-test-reports</module>
+
         <module>legend-pure-m4</module>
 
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -30,6 +30,8 @@
     <packaging>pom</packaging>
 
     <modules>
+        <module>legend-pure-test-reports</module>
```
ivan-kyosev-gs, line 33: The module order is being changed to address an issue with the nexus staging plugin.
If the last module has a configuration to skip it's deployment, then the deployment of the whole project is skipped:

https://issues.sonatype.org/browse/NEXUS-9138

PR #431, opened 2022-01-26
Improve support for GraphQL introspection. Add support for Jackson an…

pierredebelen:
…notation in Protocol.


dave-wathen (commented):
Java import changes look good

PR #432, opened 2022-01-30
Fixes for protobuf generation to include syntax and import Timestamp proto

jackie-h:
Verified that the output generates properly using protoc.


jackie-h:
Will reraise with CLA now approved

PR #454, opened 2022-02-14
Bug fixes for Sonar

kevin-m-knight-gs:
Fix a bug in the start node path regex detected by Sonar.

Also, explicitly throw NoSuchElementException in PackageTreeIterable so that Sonar will no longer complain. This would have been thrown already, though via a call to Deque.removeFirst.

Also, simplify some logic in IncrementalCompiler_New so that (hopefully) Sonar will understand when repoTransaction is not null.

### DIFF
--- a/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/tools/PackageTreeIterable.java
+++ b/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/tools/PackageTreeIterable.java
@@ -28,6 +28,7 @@ import org.finos.legend.pure.m4.coreinstance.CoreInstance;
 import java.util.ArrayDeque;
 import java.util.Deque;
 import java.util.Iterator;
+import java.util.NoSuchElementException;
 import java.util.Spliterator;
 import java.util.Spliterators;
 import java.util.function.Consumer;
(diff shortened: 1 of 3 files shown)


legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/tools/PackageTreeIterable.java
```diff
@@ -138,7 +139,11 @@ public boolean hasNext()
         @Override
         public Package next()
         {
-            Package pkg = this.deque.removeFirst();
+            Package pkg = this.deque.pollFirst();
```
akphi, line 142: Im wondering why Sonar is complaining about this? Isnt this exactly the difference between `removeFirst` vs `pollFirst`?

akphi (approved):

legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/tools/PackageTreeIterable.java
kevin-m-knight-gs: I wondered the same thing. But I figure it's just as well to have the throw be explicit.

PR #460, opened 2022-02-21
Add GCP Workload Identity Federation Auth Strategy

abhishoya-gs:


legend-pure-code-compiled-core-relational/src/main/resources/core_relational/relational/protocols/pure/vX_X_X/models/metamodel_connection.pure
```diff
@@ -55,6 +55,18 @@ Class meta::protocols::pure::vX_X_X::metamodel::store::relational::connection::a
 {
 }
 
+Class meta::protocols::pure::vX_X_X::metamodel::store::relational::connection::alloy::authentication::GCPWorkloadIdentityFederationAuthenticationStrategy extends meta::protocols::pure::vX_X_X::metamodel::store::relational::connection::alloy::authentication::AuthenticationStrategy
+{
+    workloadProjectNumber : String[1];
+    serviceAccountEmail : String[1];
+    gcpScope: String[1];
```
epsstan, line 62: Would we ever need to include more than one scope ?

Given that the metamodel is immutable, I'd rather that we allow multiple scopes in the metamodel.

epsstan: Say we need only a single scope for BigQuery. If all specifications have to specify the same scope value, should we simply hard code it in Engine ? (Think your first implementation did this).

So how about something like this - "Standard" scopes that are needed for BigQuery are automatically provided by Engine. If there is a need, users can specify one/more scopes through the model spec. In engine we can validate/uniqueify the scopes as needed.

legend-pure-code-compiled-core-relational/src/main/resources/core_relational/relational/protocols/pure/vX_X_X/models/metamodel_connection.pure
```diff
@@ -55,6 +55,18 @@ Class meta::protocols::pure::vX_X_X::metamodel::store::relational::connection::a
...
+    workloadPoolId : String[1];
+    workloadProviderId: String[1];
+    discoveryUrl: String[1];
+    clientId: String[1];
```
epsstan, line 66: Should this be broken down into 2 separate strategies ?  One that uses AWS as the IDP and one that uses Ping/other as the IDP.

'discoveryUrl'/'clientId' make sense only for IDPs that require them. So why ask users of AWS IDP to provide dummy values for these attributes ?

PR #474, opened 2022-03-02
Export default literal values for pk cols in union dataflow

vikask-gs:
When loading result of a union query into a table, we need to define a pk. That pk can be defined as union of pks of individual tables within union. Problem here is that null isn't allowed in pk, but union query exports null for pk cols which are not present in source tables. The fix in this PR is to export default values instead of null, when importDataFlow is enabled.


pierredebelen:
Isn't 'null' ok as I would not expect the values to be used?

vikask-gs:
> Isn't 'null' ok as I would not expect the values to be used?

For different pk col types, we need different null values. 'null' would be ok for string, but for int we would need 0, for dates some default date value and so on.

PR #476, opened 2022-03-02
Implement CNAS persistence context

davidsunglee:
Extensions to persistence model for Cloud Native Alloy Streaming, including logic for file generation.


legend-pure-code-compiled-core/src/main/resources/core/external/format/cpb/tests/testCpbGenerator.pure
```diff
@@ -0,0 +1,88 @@
...
+        )
+      )
+    )
+  );
+
+  $pp;
+}
+
+
+function <<test.Test>> {serverVersion.start='v1_20_0'} meta::external::format::cpb::generation::tests::transform_persistencePipeToBackend(): Boolean[1]
```
rafaelbey, line 81: I think you want the next protocol version here, since your change are added to x_x_x, which will become that new version once it is frozen/released.

PR #503, opened 2022-03-25
Protobuf format module

jackie-h:


pierredebelen:
Shouldn't we move all the protobuf code to the module?

jackie-h:
Yes, but I'd like to be done in a separate commit. It became rather involved due to all the versions of services/pieces. Not sure if we have to keep all of those.
This initial commit enables allows us to extend meta::external::shared::format::metamodel::SchemaDetail, which can't be done without the separate module.
WDYT ?

jackie-h:
Protobuf code moved to engine, so this no longer applies to Pure. Closing

PR #507, opened 2022-03-28
Databricks PURE client versions

aamend:
Back processing previous PURE client versions for databricks integration. As SDLC maven plugin uses production version by default (currently set to `v1_22_0`), pure models that include APISpecs or DatabricksDatasource could not be compiled.

I see `v1_23_0` is now supported, though not reflected back on legend engine yet (see issue https://github.com/finos/legend-engine/issues/620). 

This change to back process `v1_21_0` and `v1_22_0` only.


epsstan:
@aamend - Protocol versions are meant to be immutable. Let's release this as part of the next protocol version. 

@hardikmaheshwari - Are we planning on a protocol release next week ?

epsstan:
Closing this PR as it is has conflicts and might not be needed. Please re-open if still relevant.

PR #508, opened 2022-03-29
WIP: correct typo for trimming column names

gs-gunjan:
need to add in tests for the whole flow


gs-gunjan:
reverting the change for now

PR #510, opened 2022-03-29
Fix: trim column names for snowflake and h2

emilia-sokol-gs:
Fix: trim column names for snowflake and h2


Yasirmod17 (approved):
Approved

PR #518, opened 2022-04-04
Tests to ensure optional parameters are handled correctly when not provided using greaterThan and LessThan filtering operations.

vj2050gs:


rafaelbey (approved):
Looks good!  Lets link the original pure PR that triggered these changes.

vj2050gs:
This was a feedback/request for the same https://github.com/finos/legend-pure/pull/478 to ensure we understand less than and greater than operations.

PR #525, opened 2022-04-08
SqlQueryToString refactoring: move literal, window and semi-structured processing to extension

vikask-gs:
This is a continuation of previously merged PR, which introduced the db extension. In this review, I am moving processing of literals, window, etc into db extension instead of being based on adhoc if else blocks on db type.


davidharte-gs (approved):
Approved

PR #526, opened 2022-04-08
Add sorting for defects addition to result

hardikmaheshwari:
Add sorting for defects addition to result


aziemchawdhary-gs (approved):
approved

PR #529, opened 2022-04-11
Set useAppliedFunction for propertyMapping and QualifiedProperty for Class

mrudula-gs:
The "useAppliedFunctions" flag in the PMCD generation is false for mappings, setting it to true.
When it is False, the function gets in-lined into whenever it's used.
When it is True, the function reference is created and used.


legend-pure-code-compiled-core/src/main/resources/core/pure/mapping/m2m/tests/simple.pure
```diff
@@ -177,8 +179,44 @@ function meta::pure::mapping::modelToModel::test::simple::testSimpleProjectAsser
...
 
+function meta::pure::mapping::modelToModel::test::pmcd::projection::getTransactionTime(src : Loans[1]) : DateTime[0..1]
+{
+   let transactionTime = $src.transactionTime;
+   if($transactionTime->isNotEmpty(),
+       |$transactionTime->toOne(),
+       | [] );
+}
 
+function <<meta::pure::profiles::test.Test>> meta::pure::mapping::modelToModel::test::pmcd::testM2MMappingPMCD(): Boolean[1]
```
davidharte-gs, line 195: ... as discussed please see if you can reuse an existing test/mapping ... thx

davidharte-gs (changes requested):

davidharte-gs (approved):

PR #550, opened 2022-04-29
refactor relational DDL , insert commands in db specific extensions

PrateekGarg-gs:
Currently there are some helper functions that generate setup sqls given input data and store definition. These are only defined for H2 . Different DBs can have different sql syntax for DDL and insert . This PR adds support for db specific ddl sql commands generation.


PrateekGarg-gs:
As base code has been moved to legend-engine now, raised  https://github.com/finos/legend-engine/pull/665

PR #600, opened 2022-07-27
Handle properties with type parameters defined on supertype

rafaelbey:
This ensure we generate proper java code when a property contains type arguments, and these type arguments are then defined on a subclass of it.

### DIFF
--- a/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/coreinstance/helper/PropertyTypeHelper.java
+++ b/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/coreinstance/helper/PropertyTypeHelper.java
@@ -1,0 +1,45 @@
+// Copyright 2022 Goldman Sachs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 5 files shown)


legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/coreinstance/helper/PropertyTypeHelper.java
```diff
@@ -0,0 +1,45 @@
...
+    {
+
+    }
+
+    public static CoreInstance getPropertyResolvedReturnType(CoreInstance classGenericType, CoreInstance property, ProcessorSupport processorSupport)
+    {
+        CoreInstance functionType = processorSupport.function_getFunctionType(property);
+        CoreInstance returnType = Instance.getValueForMetaPropertyToOneResolved(functionType, M3Properties.returnType, processorSupport);
+        CoreInstance propertyOwner = functionType.getValueForMetaPropertyToMany(M3Properties.parameters).getFirst().getValueForMetaPropertyToOne(M3Properties.genericType);
+        if (!GenericType.isGenericTypeFullyConcrete(returnType, true, processorSupport)
```
rafaelbey, line 37: While removing duplication, this line its the only change.

The current implementation uses: `!GenericType.isGenericTypeConcrete(returnType, processorSupport)`

This only checks if the rawType its a type argument, and there are scenarios where the type its concreate, but this reference type arguments.  

Moving to this API that its recursive on checking type arguments.

pierredebelen (approved):

PR #606, opened 2022-08-16
Pure IDE Runtime Options Panel

aziemchawdhary-gs:
Add a panel to Pure IDE Light to enable various options used for execution plan development and debugging.

This change also supports setting options via system properties, as well as a GUI panel


aziemchawdhary-gs:
Screenshot to show what new panel looks like

Screenshot_20220816_165544

PR #608, opened 2022-09-12
Update for compiled mode so that evaluation of lambdas where the expressionSequence has modified is in sync / matches

aormerod-gs:
Previously modifying the expression sequence and then invoking eval caused the (original pre-compiled) pureFunction, with logic corresponding to the original expression sequence, to executed giving incorrect / unexpected results.

Primary change is to make PureCompiledLambda immutable.  It is essentially a wrapper around a LambdaFunction that pairs it with a pre-compiled / generated implementation of SharedPureFunction.  This change modifies this wrapper class so that on any modifications it unwraps / returns the LambdaFunction, which will then be routed to DynamicPureFunctionImpl to be handled dynamically.

### DIFF
--- a/legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/JavaSourceCodeGenerator.java
+++ b/legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/JavaSourceCodeGenerator.java
@@ -58,8 +58,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 
-public final class JavaSourceCodeGenerator
-{
+public final class JavaSourceCodeGenerator {
     public static final String EXTERNAL_FUNCTIONS_CLASS_NAME = "PureExternal";
 
     public static final String imports =
@@ -153,8 +152,7 @@ public final class JavaSourceCodeGenerator
 
     private final String name;
 
-    public JavaSourceCodeGenerator(ProcessorSupport processorSupport, IdBuilder idBuilder, CodeStorage codeStorage, boolean writeFilesToDisk, Path directoryToWriteFilesTo, boolean includePureStackTrace, Iterable<? extends CompiledExtension> extensions, String name, String externalAPIPackage, boolean generateCompilerExtensionCode)
-    {
+    public JavaSourceCodeGenerator(ProcessorSupport processorSupport, IdBuilder idBuilder, CodeStorage codeStorage, boolean writeFilesToDisk, Path directoryToWriteFilesTo, boolean includePureStackTrace, Iterable<? extends CompiledExtension> extensions, String name, String externalAPIPackage, boolean generateCompilerExtensionCode) {
         this.name = name;
         this.processorSupport = processorSupport;
         this.idBuilder = (idBuilder == null) ? IdBuilder.newIdBuilder(this.processorSupport) : idBuilder;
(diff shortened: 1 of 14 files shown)


aormerod-gs:
Pre-commit builds completed

legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/processors/support/DynamicPureFunctionImpl.java
```diff
@@ -93,154 +83,138 @@ public T execute(ListIterable<?> vars, ExecutionSupport es)
...
-    public static PureLambdaFunction<Object> createPureLambdaFunction(LambdaFunction<?> func, MutableMap<String, Object> openVariables, Bridge bridge)
-    {
-        return createPureLambdaFunctionWrapper(new DynamicPureLambdaFunctionImpl<>(func, openVariables, bridge));
+    public static SharedPureFunction<Object> createPureFunction(FunctionDefinition<?> func, MutableMap<String, Object> openVariables, Bridge bridge) {
+        DynamicPureFunctionImpl<Object> impl = new DynamicPureFunctionImpl<>(func, openVariables, bridge);
+        if (func instanceof LambdaFunction) {
+            return createPureLambdaFunctionWrapper(impl);
+        } else {
+            return impl;
+        }
```
kevin-m-knight-gs, line 107:
```suggestion
    public static SharedPureFunction<Object> createPureFunction(FunctionDefinition<?> func, MutableMap<String, Object> openVariables, Bridge bridge)
    {
        DynamicPureFunctionImpl<Object> impl = new DynamicPureFunctionImpl<>(func, openVariables, bridge);
        return (func instanceof LambdaFunction) ? createPureLambdaFunctionWrapper(impl) : impl;
```

legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/processors/support/Pure.java
```diff
@@ -461,23 +461,27 @@ public static Object evaluate(ExecutionSupport es, org.finos.legend.pure.m3.core
...
+                        paramInstances.add(CompiledSupport.toOne(((RichIterable<?>) instances[i]), null));
+                    }
+                    else
+                    {
+                        paramInstances.add(instances[i]);
+                    }
                 }
+                return p.execute(paramInstances, es);
+            } else {
+                pureFunc.execute(Lists.fixedSize.with(instances), es);
```
kevin-m-knight-gs, line 483: Don't you still need the same type of parameter checking as above (i.e., lines 472-479)?

legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/processors/support/Pure.java
```diff
@@ -542,13 +542,9 @@ public static Object _evaluateToMany(ExecutionSupport es, Bridge bridge, org.fin
...
             }
-            if (func instanceof ConcreteFunctionDefinition)
+            if (func instanceof NativeFunction || func instanceof LambdaFunction || func instanceof ConcreteFunctionDefinition)
             {
-                return CompiledSupport.executeFunction(func, paramClasses, paramInstances, es);
-            }
-            if (func instanceof NativeFunction || func instanceof LambdaFunction)
-            {
-                SharedPureFunction<?> foundFunc = getNativeOrLambdaFunction(es, func);
+                SharedPureFunction<?> foundFunc = findSharedPureFunction(func, bridge, es);
```
kevin-m-knight-gs, line 547: Does findSharedPureFunction handle NativeFunction properly?

legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/JavaSourceCodeGenerator.java
```diff
@@ -532,10 +460,10 @@ private String buildPureCompiledLambda(ProcessorContext processorContext)
...
                 "    public PureCompiledLambda copy()\n" +
                 "    {\n" +
-                "        PureCompiledLambda p =  new PureCompiledLambda();\n" +
-                "        p.lambdaFunction = (LambdaFunction)((AbstractCoreInstance)this.lambdaFunction).copy();\n" +
-                "        p.pureFunction = this.pureFunction;\n" +
-                "        return p;\n" +
+                "        return new PureCompiledLambda(" +
+                "                   (LambdaFunction)((AbstractCoreInstance)this.lambdaFunction).copy(),\n" +
+                "                   this.pureFunction" +
+                "                   );\n" +
```
kevin-m-knight-gs, line 466: Adding newlines to go with the indentation.
```suggestion
                "        return new PureCompiledLambda(\n" +
                "                   (LambdaFunction)((AbstractCoreInstance)this.lambdaFunction).copy(),\n" +
                "                   this.pureFunction\n" +
                "                   );\n" +
```

kevin-m-knight-gs (changes requested):
A few questions and suggestions. Most of the suggestions are minor in nature.

legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/processors/support/Pure.java
aormerod-gs: Yes

Lamba functions are handled at https://github.com/finos/legend-pure/blob/3965e4e69af087750b5d9f6dcec8b8ef35863538/legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/processors/support/Pure.java#L385

Native functions are handled at https://github.com/finos/legend-pure/blob/3965e4e69af087750b5d9f6dcec8b8ef35863538/legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/processors/support/Pure.java#L434 which is the same logic as at https://github.com/finos/legend-pure/blob/3965e4e69af087750b5d9f6dcec8b8ef35863538/legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/processors/support/Pure.java#L583

legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/processors/support/Pure.java
aormerod-gs: I think this conversion happens within the SharedPureFunction implementation, but will double check

aormerod-gs: The parameter adapting seems specific to JavaMethodWithParamsSharedPureFunction (I assume because it needs to be fully type correct for invoking the method where others can handle item or [item]).  Perhaps in a future change it should be moved in to the JavaMethodWithParamsSharedPureFunction execute implementation.

kevin-m-knight-gs (approved):

PR #641, opened 2023-01-15
Pure IDE: more improvements

akphi:
Refer to https://github.com/finos/legend-studio/pull/1822 and https://github.com/finos/legend-studio/pull/1877 for the full changelog

- [x] Make web application URL configurable
- [x] `REQUIRES BACKEND` Better indicator/visual cue for readonly files; disabling edition of readonly sources
- [x] Hotkeys: `Alt + Z` for toggling text-wrap in file editor
- [x] Hotkeys: Support multiple hotkeys for find files/text commands to make it more similar to `vscode`
- [x] Upgrade `monaco-editor` to DEV version to fix problem with bracket matching: this is temporary solution, we will need to ugrade to stable version when available
- [x] `REQUIRES BACKEND` Fix `Ctrl+B` on function descriptor doesn't work, e.g. `contains_Any_MANY__Any_1__Boolean_1_`
- [x] Allow fixing compilation on very first start (in case initialization failed)
- [x] Clear compilation errors in all file before `executeGo`: right now, if you make mistake in one place and go to another file to fix it, for example, delete a duplicated function defined in another file, the compilation does not go away in the other file
- [x] Improve syntax highlighting:
  - [x] Improve on syntax highlighting of function descriptor `execution_StoreQuery_1__RoutedValueSpecification_$0_1$__Mapping_1__Runtime_1__ExecutionContext_1__Extension_MANY__DebugContext_1__Result_1_`: should we consider adding `$` to the definition of word
  - [x] Support syntax highlight for string escaping: e.g. `'Finding all instances of class \''`
  - [x] Fix syntax-highlighting for long generic types: `paths: Pair<String, PathInformation>[*], connection : DatabaseConnection[1]` will look weird: we probably should do it in a way that `<` push and `>` pops the context (watch out for arrows), a few cases to test `Person->patchByIds(^Map<String, Any>()->cast(@Map<String, Any>[*])->put('children', []), ['1']);`, `Function<{T[1]->Boolean[1]}>`
- [x] Allow cleaning test result
- [x] Embedded a terminal inside of console using `xterm`
  - [x] Adopt VSCode theme
  - [x] Clear console
  - [x] Toggle preserve log
  - [x] Search bar
  - [x] Navigatable error source information in console: e.g. `resource:/welcome.pure line:22 column:3`
  - [x] Implement `help` command and integrate extension mechanism for that, and `ansi` command
  - [x] Basic commands: `clear`, `help`, `ansi`, `go`, etc.
  - [x] Implement word navigation using keyboard and backspace/delete
  - [x] Support history of commands
  - [x] Implement `echo`
  - [x] Implement `cd`, `ls` (40mins each) - no slash -> not a path, must be a concept
  - [x] Implement `code`/`open`/`edit`/`vi` - no slash -> not a path, must be a concept
  - [x] Implement `welcome` command to open `welcome.pure` file
  - [x] Fixed a minor issue with cursor movement at extremes (one problem remains - https://github.com/xtermjs/xterm.js/issues/4405)
- [x] Bug fixes:
  - [x] Fix suggestion for copyright not working, when typing `/copyright`: in general, we now support suggestions for special commands, starting with `/`
  - [x] Fix toggling expansion caret not working in test runner panel
- [x] Componentize the test result console into single component to be used in test result panel so as to allow clicking on links
- [x] Reworked search panel
  - [x] split search panel into `search`, `reference` and `suggestion`:
  - [x] only show suggestion when suggestion state is available
  - [x] setup `References` panel to have refresh, clear, placeholder message, collapse
  - [x] redefine hotkey behavior, move search popup to inside panel, follow VSCode behavior (handle Enter key and debounced search)

### DIFF
--- a/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/filesystem/PureCodeStorage.java
+++ b/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/filesystem/PureCodeStorage.java
@@ -444,10 +444,21 @@ public class PureCodeStorage implements MutableCodeStorage
     }
 
     @Override
-    public boolean isImmutable(String path)
+    public CodeRepository getRepositoryForPath(String path)
     {
-        RepositoryCodeStorage codeStorage = getCodeStorage(path);
-        return codeStorage instanceof ImmutableRepositoryCodeStorage;
+        if (!path.isEmpty() && (path.charAt(0) == '/'))
+        {
+            int index = path.indexOf('/', 1);
+            String rootPath = index != -1 ? path.substring(1, index) : path.substring(1);
+            return this.repositoriesByName.get(rootPath);
+        }
+        return null;
+    }
+
+    @Override
+    public boolean isRepositoryImmutable(CodeRepository repository)
+    {
+        return repository != null && this.codeStorageByName.get(repository.getName()) instanceof ImmutableRepositoryCodeStorage;
     }
 
     @Override
@@ -1022,7 +1033,6 @@ public class PureCodeStorage implements MutableCodeStorage
             return CodeStorageNodeStatus.UNKNOWN;
         }
     }
-
 
     public static String getSourceRepoName(String sourceId)
     {
--- a/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/filesystem/PureCodeStorage.java
+++ b/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/filesystem/PureCodeStorage.java
@@ -41,6 +41,7 @@ import org.finos.legend.pure.m3.serialization.filesystem.repository.ScratchCodeR
 import org.finos.legend.pure.m3.serialization.filesystem.usercodestorage.CodeStorageNode;
 import org.finos.legend.pure.m3.serialization.filesystem.usercodestorage.CodeStorageNodeStatus;
 import org.finos.legend.pure.m3.serialization.filesystem.usercodestorage.CodeStorageTools;
+import org.finos.legend.pure.m3.serialization.filesystem.usercodestorage.ImmutableRepositoryCodeStorage;
 import org.finos.legend.pure.m3.serialization.filesystem.usercodestorage.MutableCodeStorage;
 import org.finos.legend.pure.m3.serialization.filesystem.usercodestorage.MutableRepositoryCodeStorage;
 import org.finos.legend.pure.m3.serialization.filesystem.usercodestorage.RepositoryCodeStorage;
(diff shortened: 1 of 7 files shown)


legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/filesystem/usercodestorage/CodeStorageTools.java
```diff
@@ -163,4 +163,18 @@ public static String joinPaths(String... paths)
...
+    public static String getRepoNameForPath(String path)
+    {
+        if (!path.isEmpty() && (path.charAt(0) == '/'))
+        {
+            int index = path.indexOf('/', 1);
+            if (index != -1)
+            {
+                return path.substring(1, index);
+            }
+            return path.substring(1);
```
kevin-m-knight-gs, line 176: This is not correct. If there is no subsequent '/', then the repo should be null. For example, "/welcome.pure" should result in null being returned.

legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/filesystem/usercodestorage/CodeStorage.java
```diff
@@ -75,6 +75,11 @@
      */
     boolean isVersioned(String path);
 
+    default boolean isRepositoryImmutable(String name)
```
kevin-m-knight-gs, line 78: I tend to think this should be on MutableCodeStorage, since it applies to questions of mutability.

legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/filesystem/PureCodeStorage.java
```diff
@@ -442,6 +443,22 @@ public boolean isVersioned(String path)
...
+        if (repo == null)
+        {
+            return false;
+        }
+        RepositoryCodeStorage repositoryCodeStorage = this.codeStorages.select(storage -> storage.getRepositories().contains(repo)).getFirst();
+        if (repositoryCodeStorage == null)
+        {
+            return false;
+        }
+        return repositoryCodeStorage instanceof ImmutableRepositoryCodeStorage;
```
kevin-m-knight-gs, line 459: Repository code storages are already indexed by repository name. Also, the null check is redundant, since the instanceof check will return false for null.
```suggestion
        return this.codeStorageByName.get(name) instanceof ImmutableRepositoryCodeStorage;
```

kevin-m-knight-gs (changes requested):

kevin-m-knight-gs (approved):

PR #670, opened 2023-02-28
Trino connector | Adding Trino as new DatabaseType

haldes:
As part of the effort to add a new Trino connector, we are adding Trino as the new Database Type.


haldes:
Closing this as it looks like Trino DatabasType ias already added.

PR #680, opened 2023-03-29
IDE Light: Read runtime options from system properties

aziemchawdhary-gs:
Allow setting of runtime options from system properties in Pure IDE Light. Allows developers to set options in execution plan generation to aid debugging/development in Pure IDE Light (for example options look [here](https://github.com/finos/legend-engine/blob/aaa58d3c160c35ce2938e65fe847c3fedbef1173/legend-engine-pure-code-compiled-core/src/main/resources/core/pure/protocol/vX_X_X/invocations/execution.pure#L187))

### DIFF
--- a/legend-pure-ide-light/src/main/java/org/finos/legend/pure/ide/light/api/RuntimeOptions.java
+++ b/legend-pure-ide-light/src/main/java/org/finos/legend/pure/ide/light/api/RuntimeOptions.java
@@ -1,0 +1,63 @@
+// Copyright 2023 Goldman Sachs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.finos.legend.pure.ide.light.api;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.swagger.annotations.Api;
+import org.finos.legend.pure.ide.light.session.PureSession;
+
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import javax.ws.rs.GET;
+import javax.ws.rs.Path;
+import javax.ws.rs.PathParam;
+import javax.ws.rs.core.Context;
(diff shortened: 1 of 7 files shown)


legend-pure-ide-light/src/main/java/org/finos/legend/pure/ide/light/session/PureSession.java
```diff
@@ -85,10 +84,27 @@ private FunctionExecution initialize()
 
         this.functionExecution = new FunctionExecutionInterpreted(VoidExecutionActivityListener.VOID_EXECUTION_ACTIVITY_LISTENER);
 
+        for(String property : System.getProperties().stringPropertyNames())
```
rafaelbey, line 87: Is not all this available already using `RuntimeOptions.systemPropertyOptions("pure.option.")`

That seems to avoid the need of extra methods, and parsing, etc.

aziemchawdhary-gs: Good point - I hadn't spotted this method.

However I've updated the PR to add an API to toggle options - the intention is that the state of runtime options is kept in PureSession.
At startup we read any options set by system properties and then the options can be toggled at runtime without having to restart the IDE.

This API could also be used by the frontend so I'm exposing it now and then the frontend can be updated later to take advantage of this.

legend-pure-ide-light/src/main/java/org/finos/legend/pure/ide/light/api/RuntimeOptions.java
```diff
@@ -0,0 +1,32 @@
...
+{
+    private PureSession pureSession;
+
+    public RuntimeOptions(PureSession pureSession)
+    {
+        this.pureSession = pureSession;
+    }
+
+    @GET
+    @Path("setOption/{name}/{value}")
```
akphi, line 20: From UI perspective, we should have an endpoint that could return the entire map so that we can populate the initial states in the web application

legend-pure-ide-light/src/main/java/org/finos/legend/pure/ide/light/session/PureSession.java
```diff
@@ -48,17 +44,26 @@
...
 
 import static org.finos.legend.pure.ide.light.helpers.response.ExceptionTranslation.buildExceptionMessage;
 
+
 public class PureSession
 {
-    private final PureRuntime pureRuntime;
+    public static final String LEGEND_TEST_SERVER_HOST = "legend.test.server.host";
+    public static final String LEGEND_TEST_SERVER_PORT = "legend.test.server.port";
+    public static final String LEGEND_TEST_H2_PORT = "legend.test.h2.port";
```
rafaelbey, line 61: I think certain choices wont work by updating this property dynamically.  Will try to share our problems

PR #687, opened 2023-04-18
Add support for byte data type

hardikmaheshwari:


legend-pure-m3-core/src/test/java/org/finos/legend/pure/m3/serialization/grammar/v1/TestByteCompilation.java
```diff
@@ -0,0 +1,64 @@
...
+                    "Class myClassWithByteProperty\n" +
+                            "{\n" +
+                            "   prop: Byte[1];\n" +
+                            "}");
+            Assert.fail("Expected a exception");
+        }
+        catch (Exception e)
+        {
+            Assert.assertEquals("Compilation error at (resource:fromString.pure line:3 column:4), \"The property 'prop' has type of 'Byte'. 'Byte' type is not supported for property.\"", e.getMessage());
+        }
```
kevin-m-knight-gs, line 62: Let's use Assert.assertThrows here.
```suggestion
        PureCompilationException e = Assert.assertThrows(PureCompilationExcetion.class, () -> compileTestSource("fromString.pure",
                    "Class myClassWithByteProperty\n" +
                            "{\n" +
                            "   prop: Byte[1];\n" +
                            "}"));
        Assert.assertEquals("Compilation error at (resource:fromString.pure line:3 column:4), \"The property 'prop' has type of 'Byte'. 'Byte' type is not supported for property.\"", e.getMessage());
```

kevin-m-knight-gs (changes requested):
One minor change request to use Assert.assertThrows for the test that checks for a compilation error.

kevin-m-knight-gs (approved):

PR #716, opened 2023-06-30
Introduce ConnectionElementAssociation

haroonsherjan:


rafaelbey (commented):
Few comments:

- Can we minimize repackaging? meta::pure<-meta::core
- Is ok to have both pkgs?  Is 'xt' the right name here?
  - import meta::relational::runtime::*;
  - import meta::xt::relational::runtime::*;

rafaelbey (approved):

PR #724, opened 2023-08-15
Assortment of new functions

gs-jp1:
reverse, cot, sign, ascii, char, log10

### DIFF
--- a/legend-pure-m2-functions-pure/src/main/resources/platform_functions/string/ascii.pure
+++ b/legend-pure-m2-functions-pure/src/main/resources/platform_functions/string/ascii.pure
@@ -1,0 +1,22 @@
+// Copyright 2023 Goldman Sachs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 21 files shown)


legend-pure-m2-functions-pure/src/main/resources/platform_functions/string/ascii.pure
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+native function meta::pure::functions::string::ascii(source:String[1]):Integer[1];
+
+function <<test.Test>> {test.excludePlatform = 'Java compiled'} meta::pure::functions::string::tests::ascii::testSimpleAscii():Boolean[1]
+{
+    assertEquals(0, ascii(''));
+    assertEquals(32, ascii(' '));
+    assertEquals(97, ascii('abc'));
```
rafaelbey, line 21: Is this correct?  What it means to get ascii code of multiple characters?  Or 97 is the ascii code of  'a' and 'bc' are ignored?

gs-jp1: Yes gets ascii code of first char. Appreciate we should add doc to the function to explain

PR #734, opened 2023-09-20
Support PureIdeLight to spin up with subset of available repositories

hardikmaheshwari:


kevin-m-knight-gs (commented):
You should look at CodeRepositorySet, particularly the subset method. I believe it's already implementing the logic you're trying to do here.

kevin-m-knight-gs (changes requested):
Two minor changes.

kevin-m-knight-gs (approved):

PR #741, opened 2023-11-06
Pure Runtime and VCS fixes

aziemchawdhary-gs:
Add update function which takes in a String revision

### DIFF
--- a/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/filesystem/usercodestorage/composite/CompositeCodeStorage.java
+++ b/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/filesystem/usercodestorage/composite/CompositeCodeStorage.java
@@ -555,6 +555,85 @@ public class CompositeCodeStorage implements MutableVersionControlledCodeStorage
     }
 
     @Override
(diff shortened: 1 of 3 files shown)


legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/filesystem/usercodestorage/composite/CompositeCodeStorage.java
```diff
@@ -554,6 +554,85 @@ else if (codeStorage != pathCodeStorage)
...
+            RepositoryCodeStorage pathCodeStorage = getCodeStorage(path);
+            if (!(pathCodeStorage instanceof MutableVersionControlledCodeStorage))
+            {
+                throw new PureCodeStorageException("Cannot commit " + paths);
+            }
+            if (codeStorage == null)
+            {
+                codeStorage = (MutableVersionControlledCodeStorage) pathCodeStorage;
+            }
+            else if (codeStorage != pathCodeStorage)
```
rafaelbey, line 581: this expect same object?  Or this should be Object.equals()?

rafaelbey (approved):

PR #768, opened 2024-02-08
Fix a bug with the implementation of map in compiled mode

kevin-m-knight-gs:
Fix a bug with the implementation of map in compiled mode.  Specifically, the implementation used when calling map via eval/evaluate didn't properly handle the case where the collection to map over had cardinality 1 or 0..1.

In passing, add better error handling in Reactivator.


kevin-m-knight-gs:
The test is curtesy of @pierredebelen

aziemchawdhary-gs (approved):

PR #769, opened 2024-02-08
Fix map (applied to one element) when used with a dynamic function re…

pierredebelen:
…turning [0..1] in compiled mode


kevin-m-knight-gs:
This duplicates #768.

PR #807, opened 2024-05-14
Relation: Add support for pivot

akphi:


akphi:
I got this pretty weird error
External reference cannot be created for instance of meta::pure::metamodel::type::generics::GenericType: @_003tbfc(4859153) instanceOf GenericType (source information: /platform_functions_relation/relation/functions/rename.pure:17c123) 
I tried to write a test for duckDB extension
so it involves duckdb-pure module
and it refers to the function ->rename (in relation)
I have included platform_functions_relation as a dependency in the definition json file
and when building the JAR maven throws that error
it's really really strange
because I also use other relation functions such as pivot groupBy

legend-pure-core/legend-pure-m3-core/src/main/antlr4/org/finos/legend/pure/m3/serialization/grammar/m3parser/antlr/core/M3CoreParser.g4
```diff
@@ -316,10 +316,10 @@ atomicExpression:
 
 columnBuilders: TILDE (oneColSpec | (BRACKET_OPEN oneColSpec(COMMA oneColSpec)* BRACKET_CLOSE))
 ;
-
-oneColSpec: identifier ((COLON (type | lambdaParam lambdaPipe) extraFunction? ))?
+oneColSpec: colSpecName (COLON (type | lambdaParam lambdaPipe) extraFunction?)?
+;
+colSpecName: identifier | STRING
```
akphi, line 321: This change is necessary to make sure we can specify column name with spaces

legend-pure-core/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/grammar/m3parser/antlr/AntlrContextToM3CoreInstance.java
```diff
@@ -1131,89 +1138,142 @@ else if (oneColSpec.type() != null)
 
             // FunctionName
             boolean isArray = ctx.columnBuilders().BRACKET_OPEN() != null;
-            List<Boolean> nonFunctions = ListIterate.collect(ctx.columnBuilders().oneColSpec(), x -> x.type() != null | x.COLON() == null).distinct();
-            if (isArray && nonFunctions.size() > 1)
+            List<RelationColumnType> relationColumnTypes = ListIterate.collect(ctx.columnBuilders().oneColSpec(), col ->
```
akphi, line 1141: This change is to improve the readability of the code, I think it's better if we clearly show what we do for each scenario (i.e. column definition, aggregate, extend), if you disagree, I can revert this part.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/relation/colSpec.pure
akphi, line 1: this file rename is to make it more consistent with your other file name (e.g. aggColSpecArray.pure, funcColSpecArray.pure ~~ they are all named exactly like the function they define)

legend-pure-functions/legend-pure-functions-relation/legend-pure-runtime-java-extension-interpreted-functions-relation/src/main/java/org/finos/legend/pure/runtime/java/extension/external/relation/interpreted/natives/shared/Shared.java
```diff
@@ -54,17 +60,67 @@ public TestTDS getTDS(CoreInstance value, ProcessorSupport processorSupport)
                 new TestTDSInterpreted(readCsv((value.getValueForMetaPropertyToOne("csv")).getName()), repository, processorSupport);
     }
 
-    public RelationType<?> getRelationType(ListIterable<? extends CoreInstance> params, int i)
+    public RelationType<?> getRelationType(ListIterable<? extends CoreInstance> params, int i, ProcessorSupport processorSupport)
```
akphi, line 63: @pierredebelen this change is arguable a hack to make `toString()` works with `Relation<Any>`, you can see I added test which asserts on the resultant TDS after pivoting, the result is of type `Relation<Any>` and when calling `toString()` it will barf as underneath, it calls `->columns()`, and `columns()` break when calling on `Relation<Any>`.
Generally, I have tried the following example of casting and it just doesn't work
```
print(#TDS
    city, country, year, treePlanted
    NYC, USA, 2011, 5000
    NYC, USA, 2000, 5000
    SAN, USA, 2000, 2000
    SAN, USA, 2011, 100
    LND, UK, 2011, 3000
    SAN, USA, 2011, 2500
    NYC, USA, 2000, 10000
    NYC, USA, 2012, 7600
    NYC, USA, 2012, 7600
  #->pivot(~[country, city], ~[sum : x | $x.treePlanted : y | $y->plus(), count : x | $x : y | $y->size()])->cast(@Relation<(year: Integer)>)->toString());
```
This will result in error
```
Execution error at (resource:/welcome.pure line:50 column:124), "Cast exception: Relation<Any> cannot be cast to Relation<(year:Integer)>"
```
So I'm not sure how we can make `cast(@Relation<(...)>)` work with `Relation<Any>`. If I miss something, please let me know. Thanks!

akphi: Reverted as we should not hack this way. Instead, we make sure `Pivot` in interpreted return the proper generic type at execution time.

legend-pure-core/legend-pure-m3-core/src/main/antlr4/org/finos/legend/pure/m3/serialization/grammar/m3parser/antlr/core/M3CoreParser.g4
```diff
@@ -316,10 +316,8 @@ atomicExpression:
 
 columnBuilders: TILDE (oneColSpec | (BRACKET_OPEN oneColSpec(COMMA oneColSpec)* BRACKET_CLOSE))
 ;
-
-oneColSpec: identifier ((COLON (type | lambdaParam lambdaPipe) extraFunction? ))?
+oneColSpec: columnName (COLON (type | lambdaParam lambdaPipe) extraFunction?)?
```
akphi, line 319: these grammar change is to make sure we can specify column name with space in it

pierredebelen (approved):

PR #813, opened 2024-05-23
Add stereotype to Database metamodel

mrudula-gs:
Add stereotype to Database metamodel


legend-pure-store/legend-pure-store-relational/legend-pure-m2-store-relational-pure/src/main/resources/platform_store_relational/grammar/relational.pure
```diff
@@ -31,6 +31,7 @@ Class meta::relational::metamodel::Database extends meta::pure::store::set::SetB
     schemas : Schema[*];
     joins : Join[*];
     filters : Filter[*];
+    stereotypes: Stereotype[*];
```
kevin-m-knight-gs, line 34: This property should already be inherited from ElementWithStereotypes (via Database -> SetBasedStore -> Store -> PackageableElement -> ModelElement -> AnnotatedElement -> ElementWithStereotypes). So there should be no need to add it directly here.

PR #827, opened 2024-06-04
Compiled mode: set default values for new instances in reactivator flow

aziemchawdhary-gs:
This fixes a bug in the compiled mode version of reactivator where evaluation of calls to new does not set default values defined on the class being instantiated

### DIFF
--- a/legend-pure-runtime/legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/processors/support/Pure.java
+++ b/legend-pure-runtime/legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/processors/support/Pure.java
@@ -33,6 +33,7 @@ import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.function.Concre
 import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.function.FunctionDefinition;
 import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.function.LambdaFunction;
 import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.function.NativeFunction;
+import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.function.property.DefaultValue;
 import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.function.property.Property;
 import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.function.property.QualifiedProperty;
 import org.finos.legend.pure.m3.coreinstance.meta.pure.metamodel.multiplicity.Multiplicity;
@@ -731,8 +732,33 @@ public class Pure
                     {
                         m.invoke(result, Lists.fixedSize.of(res));
                     }
+
                 }
             });
+
+            // Set default values
+            aClass._properties().forEach(new CheckedProcedure<Property<?, ?>>()
+            {
+                @Override
+                public void safeValue(Property<?, ?> p) throws Exception
+                {
+                    DefaultValue defaultValue = p._defaultValue();
+                    if (defaultValue != null)
+                    {
+                        Object result = reactivate(defaultValue._functionDefinition()._expressionSequence().getFirst(), new PureMap(Maps.fixedSize.empty()), bridge, es);
+                        Method method = c.getMethod("_" + p._name(), RichIterable.class);
+                        if (result instanceof RichIterable)
+                        {
+                            method.invoke(result, result);
+                        }
+                        else
+                        {
+                            method.invoke(result, Lists.fixedSize.of(result));
+                        }
+                    }
+                }
+            });
+
 
             return result;
         }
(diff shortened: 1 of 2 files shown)


legend-pure-core/legend-pure-m3-core/src/test/java/org/finos/legend/pure/m3/tests/elements/property/AbstractTestDefaultValue.java
```diff
@@ -126,5 +127,24 @@ public void testDefaultValueWithQualifiedProperty()
...
+                    "defaultValueSource.pure",
+                     "Class test::A { i:Integer[1]; }\n" +
+                            "Class test::B { j:Integer[1]; }\n" +
+                            "Association test::AB { a : test::A[1]; b : test::B[1] = ^test::B(j=2);}"
+            );
+        }
+        catch (RuntimeException e)
+        {
+            assertPureException(PureCompilationException.class, "Association properties should not have default values defined. Error in property: b", "defaultValueSource.pure", 3, 19, 3, 71, e);
+        }
```
kevin-m-knight-gs, line 147: As this test stands, it will succeed if an exception is NOT thrown. Let's use Assert.assertThrows here.
```suggestion
        PureCompilationException e = Assert.assertThrows(PureCompilationException.class, () -> compileTestSource(
                    "defaultValueSource.pure",
                     "Class test::A { i:Integer[1]; }\n" +
                            "Class test::B { j:Integer[1]; }\n" +
                            "Association test::AB { a : test::A[1]; b : test::B[1] = ^test::B(j=2);}"
            ));
        assertPureException(PureCompilationException.class, "Association properties should not have default values defined. Error in property: b", "defaultValueSource.pure", 3, 19, 3, 71, e);
```

legend-pure-core/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/compiler/postprocessing/processor/AssociationProcessor.java
```diff
@@ -111,6 +111,16 @@ public static Iterable<AbstractProperty<?>> process(Association association, Con
...
 
+        if (leftProperty._defaultValue() != null)
+        {
+            throw new PureCompilationException(association.getSourceInformation(), "Association properties should not have default values defined. Error in property: " + leftProperty.getName());
+        }
+
+        if (rightProperty._defaultValue() != null)
+        {
+            throw new PureCompilationException(association.getSourceInformation(), "Association properties should not have default values defined. Error in property: " + rightProperty.getName());
+        }
```
kevin-m-knight-gs, line 122: I think these checks more properly belong in AssociationValidator than in AssociationProcessor. Also, we should probably use the source information of the property, rather than the association.

legend-pure-runtime/legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/processors/support/Pure.java
```diff
@@ -716,6 +718,33 @@ public static Object newObject(Bridge bridge, org.finos.legend.pure.m3.coreinsta
...
+                                method.invoke(result, res);
+                            }
+                            else
+                            {
+                                method.invoke(result, Lists.fixedSize.of(res));
+                            }
+                        }
+                    }
+                }
+            });
```
kevin-m-knight-gs, line 747: I think this block should go after the key expressions, and we should only invoke the default value if the value isn't set by a key expression. That means we'd need to keep track of which properties we've set from key expressions, but that's easy enough.

Also, I think we should just cast to Property<?, ?> here and not bother with the if check. After all, if it's not an instance of Property, we want to throw instead of just ignoring it.

legend-pure-runtime/legend-pure-runtime-java-engine-compiled/src/main/java/org/finos/legend/pure/runtime/java/compiled/generation/processors/support/Pure.java
```diff
@@ -736,9 +742,38 @@ public void safeValue(KeyExpression o) throws Exception
...
+                            if (res instanceof RichIterable)
+                            {
+                                method.invoke(result, res);
+                            }
+                            else
+                            {
+                                method.invoke(result, Lists.fixedSize.of(res));
+                            }
+                        }
+                    }
```
kevin-m-knight-gs, line 772: We have methods now on AbstractCompiledCoreInstance for dealing with default values. Maybe result above can be cast directly to AbstractCompiledCoreInstance? In case it needs to be cast to Any above, I'm casting it in the suggestion below.
```suggestion
            AbstractCompiledCoreInstance instance = (AbstractCompiledCoreInstance) result;
            instance.getDefaultValueKeys().forEach(new CheckedProcedure<String>()
            {
                @Override
                public void safeValue(String p) throws Exception
                {
                    if (!userAssignedParameters.contains(p))
                    {
                        RichIterable<?> defaultValues = instance.getDefaultValue(p, es);
                        if ((defaultValues != null) && defaultValues.notEmpty())
                        {
                            Method method = c.getMethod("_" + prop._name(), RichIterable.class);
                            method.invoke(result, defaultValues);
                        }
                    }
```

PR #832, opened 2024-06-16
relation: add support pivot()

akphi:


akphi:
Close in favor of #843 and https://github.com/finos/legend-engine/pull/2979

PR #888, opened 2024-11-13
handle stacktrace error for non-functions

gs-rpant1729:
Handle error for functions without descriptors (properties, qualified properties etc.). In the absence of this change, when using Pure IDE Light, you get a full server-side Java stacktrace (instead of a Pure one).

### DIFF
--- a/legend-pure-core/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/exception/PureExecutionException.java
+++ b/legend-pure-core/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/exception/PureExecutionException.java
@@ -137,7 +137,14 @@ public class PureExecutionException extends PureException
                 CoreInstance func = x.getValueForMetaPropertyToOne(M3Properties.func);
                 if (func != null)
                 {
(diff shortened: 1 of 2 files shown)


legend-pure-core/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/exception/PureExecutionException.java
```diff
@@ -137,7 +137,14 @@ public <T extends Appendable> T printPureStackTrace(T appendable, String indent,
...
                 {
-                    FunctionDescriptor.writeFunctionDescriptor(appendable, func, false, processorSupport);
+                    try
+                    {
+                        FunctionDescriptor.writeFunctionDescriptor(appendable, func, false, processorSupport);
+                    }
+                    catch (IllegalArgumentException e)
+                    {
+                        safeAppendable.append("NULL / TODO");
+                    }
```
kevin-m-knight-gs, line 147: We should be able to identify and handle properties, rather than just writing "NULL / TODO".

gs-rpant1729:
Already handled in https://github.com/finos/legend-pure/pull/889.

PR #929, opened 2025-02-07
Fix type inference on function processing for relation

gs-jp1:
Currently for looking up properties where the var type is a relation, it is trying to find the property on type Relation and not the actual relation type


gs-jp1:
Closing PR - under consideration I dont think this is a correct change to make, it effectively makes any collection function work at the row level of a the relation which is not inline with expectation.

PR #939, opened 2025-03-03
adding code for time_bucket native function

mimi-tam:
adding code for time_bucket native function

### DIFF
--- a/legend-pure-core/legend-pure-m4/src/main/java/org/finos/legend/pure/m4/coreinstance/primitive/date/PureDate.java
+++ b/legend-pure-core/legend-pure-m4/src/main/java/org/finos/legend/pure/m4/coreinstance/primitive/date/PureDate.java
@@ -14,6 +14,7 @@
 
 package org.finos.legend.pure.m4.coreinstance.primitive.date;
 
+import java.time.Instant;
 import java.util.GregorianCalendar;
 
 public interface PureDate extends Comparable<PureDate>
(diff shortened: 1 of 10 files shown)


AFine-gs (approved):

legend-pure-core/legend-pure-m4/src/main/java/org/finos/legend/pure/m4/coreinstance/primitive/date/PureDate.java
```diff
@@ -77,6 +78,11 @@ public interface PureDate extends Comparable<PureDate>
      */
     GregorianCalendar getCalendar();
 
+    default Instant toInstant()
+    {
+        return this.getCalendar().toZonedDateTime().toInstant();
+    }
+
```
kevin-m-knight-gs, line 85: Not all PureDates can be converted to Instants in a valid way. Really, only those with nanosecond or better precision can. Is this method really necessary on this interface?

If it is really necessary, it should be accompanied by another method to test whether the PureDate can validly be converted to an Instant. And the default method should throw UnsupportedOperationException if it can't.

kevin-m-knight-gs (changes requested):

legend-pure-core/legend-pure-m4/src/main/java/org/finos/legend/pure/m4/coreinstance/primitive/date/PureDate.java
mimi-tam: addressed in another PR - https://github.com/finos/legend-pure/pull/943/files ; closing this PR

PR #970, opened 2025-04-06
Fix SimpleDateFormat issue

mikkoi:
Do not use SimpleDateFormat format "z"

java.test.SimpleDateFormat does not have clear rules for format "z".
The rule:
z       Time zone       General time zone       Pacific Standard Time; PST; GMT-08:00
General time zone: Time zones are interpreted as text if they have names

https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/text/SimpleDateFormat.html#timezone

Avoid using SimpleDateFormat format "z" when not necessary.

Example of errors detected:
[ERROR] Failures:
[ERROR]   TestPureDate.testFormatWithTimeZoneShift:52 expected:<...-12-31 19:01:01.070 [CST]> but was:<...-12-31 19:01:01.070 [GMT-06:00]>


legend-pure-core/legend-pure-m4/src/test/java/org/finos/legend/pure/m4/coreinstance/primitive/date/TestPureDate.java
```diff
@@ -55,6 +55,25 @@ public void testFormatWithTimeZoneShift()
...
+        Assert.assertEquals("2014-01-01 00:01:01.070 Atlantic/Azores", date.format("[Atlantic/Azores]yyyy-MM-dd HH:mm:ss.SSS z"));
+        Assert.assertEquals("2014-01-01 00:01:01.070-01", date.format("[Atlantic/Azores]yyyy-MM-dd HH:mm:ss.SSSX"));
+
+        // Time zone is illegal after date or time has been formatted and written out
+        Assert.assertEquals("Time zone can only be set at the beginning of the format string",
+                Assert.assertThrows(IllegalArgumentException.class, () -> date.format("yyyy-MM-dd HH:mm:ss.SSS [CET]z")).getMessage());
+        Assert.assertEquals("Time zone can only be set at the beginning of the format string",
+                Assert.assertThrows(IllegalArgumentException.class, () -> date.format("yyyy-MM-dd [CET]HH:mm:ss.SSSz")).getMessage());
+        Assert.assertEquals("Time zone can only be set at the beginning of the format string",
+                Assert.assertThrows(IllegalArgumentException.class, () -> date.format("yyyy-MM-dd[CET] HH:mm:ss.SSSz")).getMessage());
```
kevin-m-knight-gs, line 76: Can you add some cases here demonstrating that multiple time zones would be illegal, regardless of whether they come at the start of the string? The invalid format string from the deleted testFormatWithMultipleTimeZones should still throw an exception (though possibly a different one). Also, a format string like "[CET] [CST] yyyy-MM-dd HH:mm:ss.SSSz" should throw an exception.

legend-pure-core/legend-pure-m4/src/main/java/org/finos/legend/pure/m4/coreinstance/primitive/date/DateFormat.java
```diff
@@ -56,18 +61,18 @@ public static <T extends Appendable> T format(T appendable, String formatString,
                             timeZoneId.append(next);
                             escaped = false;
                         }
-                        else if (inQuotes)
+                        else if (inQuotes && next != '"')
```
kevin-m-knight-gs, line 64: Perhaps we should just move the next == '"' case before this case?

legend-pure-core/legend-pure-m4/src/test/java/org/finos/legend/pure/m4/coreinstance/primitive/date/TestPureDate.java
mikkoi: I made this assumption:

    if (i > 1)
                    {
                        throw new IllegalArgumentException("Time zone can only be set at the beginning of the format string");
                    }

If time zone is not set at the very beginning of the string, it simply fails.

Perhaps this is too rigid?
It doesn't accept preceeding whitespace.

kevin-m-knight-gs: I thought about that, but I think it's fine to make it that rigid. It certainly makes the logic easier and the requirement clearer.

Nonetheless, I think we should still have test cases asserting that multiple time zones will result in an exception. Even if it seems like it's implied by other asserts, I think it's good to have test cases that explicitly assert it.

legend-pure-core/legend-pure-m4/src/main/java/org/finos/legend/pure/m4/coreinstance/primitive/date/DateFormat.java
mikkoi: Fixed in commit 98bbe1bc1

legend-pure-core/legend-pure-m4/src/test/java/org/finos/legend/pure/m4/coreinstance/primitive/date/TestPureDate.java
mikkoi: Returned in commit 98bbe1bc1.

legend-pure-core/legend-pure-m4/src/main/java/org/finos/legend/pure/m4/coreinstance/primitive/date/DateFormat.java
```diff
@@ -56,18 +61,14 @@ public static <T extends Appendable> T format(T appendable, String formatString,
                             timeZoneId.append(next);
                             escaped = false;
                         }
-                        else if (inQuotes)
+                        else if (next == '"')
                         {
-                            timeZoneId.append(next);
+                            inQuotes = !inQuotes;
                         }
                         else if (next == ']')
```
kevin-m-knight-gs, line 68: Since you're eliminating the inQuotes condition, I think this condition needs an extra conjunct so that we're not ending while we're still in quotes.
```suggestion
                        else if ((next == ']') && !inQuotes)
```

kevin-m-knight-gs (approved):

PR #1023, opened 2025-06-12
Support for nulls on primitives when parsing TDS

rafaelbey:
Allow primitive types to be null. For int/long, using largest primes as unlikely we would see these on our test cases, rather than using min int/long, as these can be used for testing boundary conditions.

### DIFF
--- a/legend-pure-dsl/legend-pure-dsl-tds/legend-pure-m2-dsl-tds-grammar/src/main/java/org/finos/legend/pure/m2/inlinedsl/tds/TDSExtension.java
+++ b/legend-pure-dsl/legend-pure-dsl-tds/legend-pure-m2-dsl-tds-grammar/src/main/java/org/finos/legend/pure/m2/inlinedsl/tds/TDSExtension.java
@@ -19,6 +19,7 @@ import io.deephaven.csv.parsers.DataType;
 import io.deephaven.csv.reading.CsvReader;
 import io.deephaven.csv.sinks.SinkFactory;
 import io.deephaven.csv.util.CsvReaderException;
+import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


legend-pure-dsl/legend-pure-dsl-tds/legend-pure-m2-dsl-tds-grammar/src/main/java/org/finos/legend/pure/m2/inlinedsl/tds/TDSExtension.java
```diff
@@ -236,17 +237,22 @@ private static String convertType(DataType dataType)
         }
     }
 
-    private static SinkFactory makeMySinkFactory()
+    public static CsvSpecs makePureCsvSpecs()
```
rafaelbey, line 240: Made this utility methods public to allow using them on Legend Engine

legend-pure-dsl/legend-pure-dsl-tds/legend-pure-m2-dsl-tds-grammar/src/main/java/org/finos/legend/pure/m2/inlinedsl/tds/TDSExtension.java
```diff
@@ -236,17 +237,22 @@ private static String convertType(DataType dataType)
...
+    {
+        return CsvSpecs.builder().nullValueLiterals(Arrays.asList("", "null")).build();
```
rafaelbey, line 242: When parsing, either empty or the literal null will be treated as null values.  This is mimicking what we already do on Engine.

pierredebelen (approved):

PR #1044, opened 2025-07-07
Refactor primitive utilities to a shared module

jake-kim1:
Refactor primitive utilities to a shared module


kevin-m-knight-gs (changes requested):
It seems to me that most (or all?) of what you are proposing to move are execution utilities. Shared execution utilities should go in legend-pure-runtime-java-engine-shared, not in legend-pure-m3-core (which is more for compiler related code).

You should also consider carefully how much is actually shared: the details of handling numbers differs between interpreted and compiled modes.

PR #1297, opened 2026-07-22
Fix eval in interpreted mode to handle upcast arguments

aziemchawdhary-gs:
In interpreted mode, Evaluate.validateValueToSignature checked an argument's
  wrapper genericType, which legitimately stays wider than the actual value
  after an upcast (e.g. cast(@SetImplementation) followed by match/parameter
  binding, which pass the wrapper through unchanged). Compiled mode dispatches
  on the runtime instance, so the same code succeeded there.

  Fall back to checking each actual value's extracted generic type when the
  wrapper check fails, mirroring Match dispatch and Cast down-cast validation.
  The change is strictly relaxing: genuinely incompatible values still throw.

### DIFF
--- a/legend-pure-runtime/legend-pure-runtime-java-engine-interpreted/src/main/java/org/finos/legend/pure/runtime/java/interpreted/natives/essentials/lang/eval/Evaluate.java
+++ b/legend-pure-runtime/legend-pure-runtime-java-engine-interpreted/src/main/java/org/finos/legend/pure/runtime/java/interpreted/natives/essentials/lang/eval/Evaluate.java
@@ -34,6 +34,7 @@ import org.finos.legend.pure.m3.navigation.ProcessorSupport;
 import org.finos.legend.pure.m3.navigation.generictype.GenericType;
 import org.finos.legend.pure.m3.navigation.generictype.match.GenericTypeMatch;
 import org.finos.legend.pure.m3.navigation.generictype.match.ParameterMatchBehavior;
+import org.finos.legend.pure.m3.navigation.measure.Measure;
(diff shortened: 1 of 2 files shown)


legend-pure-runtime/legend-pure-runtime-java-engine-interpreted/src/main/java/org/finos/legend/pure/runtime/java/interpreted/natives/essentials/lang/eval/Evaluate.java
```diff
@@ -122,4 +124,23 @@ public static void validateValueToSignature(CoreInstance value, CoreInstance sig
...
+    /**
+     * The declared genericType of a value specification may be wider than the types of the actual
+     * values it holds (e.g. after an upcast, whose result keeps the cast target type on the wrapper).
+     * Dynamic evaluation must accept any value whose actual type conforms to the signature, as
+     * compiled mode dispatches on the runtime instance.
+     */
+    private static boolean actualValuesMatch(CoreInstance value, CoreInstance signatureGenericType, ProcessorSupport processorSupport)
+    {
+        ListIterable<? extends CoreInstance> values = Instance.getValueForMetaPropertyToManyResolved(value, M3Properties.values, processorSupport);
+        return values.allSatisfy(v -> GenericTypeMatch.genericTypeMatches(signatureGenericType, extractActualGenericType(v, processorSupport), true, ParameterMatchBehavior.MATCH_ANYTHING, ParameterMatchBehavior.MATCH_CAUTIOUSLY, processorSupport));
```
gs-rpant1729, line 137: Would we need a null check here? Not sure if all values will be wrapped in an InstanceValue.

PR #1299, opened 2026-07-28
Add documentation ('''...''') as sugar over the doc.doc tagged value

rafaelbey:
## Summary

A `'''...'''` literal immediately preceding a declaration parses to exactly the
`meta::pure::profiles::doc` `doc` tagged value it replaces.

```pure
'''
A **person** in the system.

Identity is established by `legalName`.
'''
Class model::Person
{
  '''
  Given name. Not guaranteed unique.
  '''
  firstName: String[1];
}
```

is identical after parsing to:

```pure
Class {meta::pure::profiles::doc.doc = 'A **person** in the system.\n\nIdentity is established by `legalName`.'} model::Person
{
  {meta::pure::profiles::doc.doc = 'Given name. Not guaranteed unique.'} firstName: String[1];
}
```

**No new profile, no M3 metamodel change, no compiler-pass change.** Consumers that already read the `doc` tag pick documentation up automatically. Content is conventionally Markdown, but the grammar is format-agnostic — nothing here parses or validates it.

## This PR changed direction — please read this section first

The first four commits added `/** ... */` documentation comments. Review feedback asked whether the existing multi-line string could carry documentation instead. It can, and the result is a **smaller** feature, so the last six commits pivot to it. The reviewed commits are left untouched, so there is nothing to re-read.

**The six new commits are sized to be read in order**, and two of them are where the substance is:

| commit | files | change | what it is |
|---|---|---|---|
| `End a source range at the real end of a multi-line token` | 2 | +136 −1 | pre-existing bug, independent of documentation |
| `Extract the multi-line layout shared by string literals and documentation` | 2 | +105 −63 | pure refactor, no behaviour change |
| `Keep multi-line tagged value content literal` | 2 | +25 −2 | small behaviour change, independently arguable |
| **`Add documentation as a '''...''' literal on declarations`** | 8 | **+885 −35** | **the new mechanism, purely additive — `/** */` still works here** |
| **`Remove '/** ... */' documentation comments`** | 11 | **+40 −974** | **the old mechanism going away, nearly all deletion** |
| `Document the '''...''' documentation form` | 5 | +116 −106 | docs and editor syntax |

Every one of them builds and passes on its own. The fourth deliberately leaves both forms working so the new mechanism can be read without the old one's removal tangled into the same diff; roughly fifteen lines of transitional code in the `taggedValues` funnel accept either, and the fifth deletes them.

Worth stating up front, because it makes one reading of the suggestion a no-op: `{meta::pure::profiles::doc.doc = '''...'''}` **already compiled on master** (`M3CoreParser.g4` accepts `MULTILINE_STRING` as a tagged-value value, pinned by `TestProfile#testMultilineTaggedValue`). What this PR adds is the bare docstring position.

What the pivot deleted outright:

| Removed | Lines |
|---|---|
| `DocCommentLookup` — hidden-channel backward scan + adjacency heuristic | 103 |
| `DocCommentCanonicalizer` — replaced by a layout shared with string literals | 193 |
| `DOCUMENTATION` channel, `DOC_COMMENT` token, recursive `DocComment` fragment | 8 |

A one-line form (`'''docs'''`) was prototyped and **deliberately dropped**: `MultilineString` requires a line terminator after the opening delimiter, so a one-line form needs either a second lexer token or a relaxation of the string literal itself, and neither is worth it here. Documentation always spans at least three lines.

## Documentation is a parser rule, not a lexer token

```antlr
documentation: MULTILINE_STRING
;
```

prefixed onto the eleven declarations that already accept `taggedValues?`.

This is deliberately **not** a second lexer token. Lexing happens before parsing, so the lexer cannot tell a documentation position from an expression position; a token matching the same text as `MULTILINE_STRING` would simply shadow it by declaration order. Listing the rule per position is exactly what leaves `'''...'''` in expression position untouched.

Two consequences, both improvements:

- **Attachment is syntactic.** No backward scan, no "whitespace containing no blank line" rule. This fixes a real wart of the previous design: a `//` note written between the documentation and its element used to detach the documentation *silently*, with no error.
- **Source information is naturally subsumed.** The literal is a child of the declaration, so the synthesized tagged value cannot fall outside the element's range — the `ReferenceIdGenerator` / PAR failure that commit 3 fixed cannot recur by construction.

The `taggedValues` funnel reads the documentation off the element context (`getRuleContext(DocumentationContext.class, 0)`) rather than taking it as a parameter, so a rule that gains `documentation?` needs no Java change.

## Content is literal, and this matters more than it did

A string literal applies `StringEscape`. Documentation must not, because it is prose. Measured by running `StringEscape.unescape` against the built classes:

| documentation content | result |
|---|---|
| `\d+` (regex) | `d+` |
| `\*not italic\*` (Markdown escape) | `*not italic*` |
| `C:\temp` | `C:` + tab |
| **`C:\users`** | **`IllegalArgumentException: Unable to parse unicode value: sers`** |

That last one is unchecked, carries no source information, and aborts compilation of the whole repository. So `MultilineTextLayout` holds the Java-text-block layout the two forms share, and `DocumentationCanonicalizer` adds blank-line trimming while stopping short of the escaping.

The same exposure existed on an **explicit** multi-line `doc.doc` tagged value, which this PR also makes literal — matching the single-line form, which has never unescaped.

## Source information for multi-line tokens (first commit, independent of the rest)

An ANTLR token records only where it *starts*: `getLine()` is documented as the line of its first character, and neither `Token` nor `CommonToken` carries an end line. `getPureSourceInformation` took the end line from that start line while taking the end column from the whole token text's length — two different coordinate systems as soon as a token spans lines. A three-line `'''...'''` reported *line 1, column 12* on a line three characters long.

Two tokens can span lines: `MULTILINE_STRING` and `DSL_TEXT`, so this reached explicit multi-line tagged values and the graph and path DSL blocks — **a pre-existing bug, present on master since multi-line strings landed**, and fixed here rather than routed around, so no caller keeps the broken behaviour.

Newlines are counted as `\n` only and the end column as the characters after the last one, which is precisely how `LexerATNSimulator.consume` advances `line` and `charPositionInLine`. For a single-line token the result is unchanged, which is why nothing asserted the old behaviour.

## Testing

| Suite | Result |
|---|---|
| **Full reactor `mvn clean install`** (JDK 17) | **107 modules, 6573 tests, 0 failures, 0 errors** |
| `TestDocumentation` (m3-core) | 45 |
| `TestDocumentationCanonicalizer` (m4) | 17 |
| `TestProfile` (m3-core) | 7 |
| `TestAntlrSourceInformation` (m4) | 4 |
| Checkstyle | 0 violations |

Grammar ambiguity was **proven, not argued**: a standalone ANTLR 4.8-1 harness reproducing the real `definition` / `properties` / `codeBlock` decision structure, run under `PredictionMode.LL_EXACT_AMBIG_DETECTION` — 26 cases, zero ambiguity reports, zero tool warnings, before the real grammars were touched. All three generating grammar modules build warning-free under `treatWarningsAsErrors`.

Two tests exist because nothing else can catch their failure mode:

- `testDocumentedElementSurvivesIncrementalRecompilation` — `processPackageableElement` needs an element's stored source start to **equal** its parse-context start on every recompile, and the four anchors involved are identical whenever an element carries no documentation, so no existing test can see a mistake there.
- `platform/pure/documentation.pure` — pins the contract at run time on both execution engines, after metadata serialization rather than read off the AST.

## Notes for reviewers

- **This is parse-only.** There is no composer in this repository, so nothing re-emits documentation; a documented element renders as an ordinary `doc.doc` tagged value wherever tagged values are printed.
- **legend-engine has its own grammar** and will reject a documented file until a matching change lands there. Its `CoreLexerGrammar.g4` skips `/* */`, so `/** */`-documented files parse there today — that is the one thing the pivot costs, and it gates rollout into `.pure` sources, not this PR.
- **The same delimiter means different things by position.** `'''...'''` before a property is documentation; inside a derived property's body it *is* the return value. The naive fix of adding `;` compiles and silently discards a string. Documented as a gotcha; adding a documentation slot inside `codeBlock` would make this worse, not better.
- **Editor highlighting is partially unfixable.** A `'''` rule has been added to the VSCode TextMate grammar (there was none, so documentation previously mis-highlighted from the first apostrophe in the prose). TextMate cannot distinguish doc position from expression position, so documentation colours as a string rather than a comment.
- The `###` section split runs before any grammar and is not string-aware, so `###` at column 0 inside documentation splits the file. Pre-existing, identical for `/** */`, documented not fixed.

## Documentation

`### Documentation` in `docs/reference/pure-language-reference.md`, rewritten rather than renamed — the adjacency, star-stripping and comment-nesting subsections are gone because the constructs are, and two new subsections cover the expression-position trap and why documentation does not process escapes. Plus `legend-grammar-reference.md`, the compiler pipeline overview, the docs index, and the VSCode syntax file.


rafaelbey:
## Lexer/parser performance and memory impact

Adding a token that is *retained* rather than skipped is a fair thing to be uneasy about, so here are measurements rather than reasoning.

### 1. The per-element lookup is O(1) with zero allocation

`DocCommentLookup` calls `getHiddenTokensToLeft` once per documentable element. That sounds costly, but ANTLR exits before doing any work in the common case:

```java
int prevOnChannel = previousTokenOnChannel(tokenIndex - 1, Lexer.DEFAULT_TOKEN_CHANNEL);
if ( prevOnChannel == tokenIndex - 1 ) return null;   // <-- exits before allocating
return filterForChannel(from, to, channel);            // <-- allocates an ArrayList
```

The reason this always hits is that whitespace and ordinary comments are `-> skip`, so they **never enter the token buffer at all**. For any element without a documentation comment, the token at `index - 1` is a real token, so the call is one array access and one comparison. `filterForChannel` — which allocates an `ArrayList` even when it ends up returning `null` — is only reached when something off-channel is genuinely present.

This is also why the existing `COMMENT` rule was deliberately left as `-> skip` rather than moved to a channel: skipped tokens cannot get in the way of the look-back.

### 2. Lexing throughput is unchanged

Two generated lexers — one with `DOC_COMMENT`, one without — over identical inputs. Each variant runs in **its own JVM**, because running both in one process shares JIT state and the static DFA cache and produces unusable numbers. 50 warmup iterations, best of 200.

| Input | Before | After | Delta |
|---|---:|---:|---:|
| No comments (889 KB) | 4.222 ms | 4.126 ms | −2.3% |
| Ordinary `/* */` (2.0 MB) | 11.314 ms | 11.269 ms | −0.4% |
| Doc `/** */` (2.0 MB) | 11.245 ms | 11.216 ms | −0.3% |

All three deltas are *negative*, which is the signature of measurement noise rather than a speedup — a real regression would be consistently positive.

The recursive non-greedy `DocComment` fragment is the part that looks expensive, but it only engages after `/**` followed by a non-`/`, so ordinary source never enters it.

### 3. Memory: ~48 bytes per documentation comment, and only during the parse

Retained heap of a filled `CommonTokenStream` (20,000 elements, each with a documentation comment):

| Variant | Tokens | Retained |
|---|---:|---:|
| Before | 160,001 | 10.33 MB |
| After | 180,001 | 11.25 MB |

+0.92 MB for +20,000 tokens = **48 bytes per documentation comment**, held only while the token stream is alive.

Checked reflectively: `CommonToken.text` is **`null`** — token text is read lazily from the `CharStream` rather than copied. So a 10 KB documentation comment occupies the same 48 bytes in the buffer as a 10-byte one. Comment size does not scale buffer memory.

### 4. Nothing new is retained long-term

The canonicalized string becomes a `TaggedValue` in the graph — which is exactly what an explicit `{meta::pure::profiles::doc.doc = '...'}` already costs today. **Net retained memory versus hand-writing the tagged value is zero**, and source with no documentation comments allocates nothing at all.

### 5. One-time cost

The extra token grows the serialized ATN: in the probe grammar the generated lexer class went 5018 → 5180 bytes (+3.2%), loaded once per classloader. `M3Lexer.class` is ~21 KB.

### Caveats

- This measures **lexing only**. Parsing, post-processing and validation dominate a real Pure build, so a ~0% lexer delta is smaller still in context; the m3-core suite (1659 tests) showed no timing change distinguishable from noise.
- The probe grammar is a small stand-in for `M3CoreLexer`'s full token set. Adding one rule to a larger DFA should be proportionally cheaper, not more expensive.
- The adjacency check does allocate — it materializes the gap substring between the comment and the element — but only for elements that actually carry a documentation comment, and the gap is a few characters.

kevin-m-knight-gs (approved):

PR #1305, opened 2026-07-31
Document every PCT function

rafaelbey:
Builds on #1299, which added the `'''…'''` documentation form. That gave every
PCT function somewhere to put documentation; this fills it in.

## What changes

**Coverage goes from 83 of 309 signatures to all 309.** Much of the previous 83
was a single plain sentence, and 16 were an empty `doc.doc=''`. Every signature
now has a summary, parameters, returns, runnable examples and cross-references.
The 31 PCT tests in `essential/string` are documented too, so they publish as
worked examples.

**PCT test documentation now reaches the report.** `FunctionDefinition` tracked
tests as two counters and nothing else, so a documentation literal on a
`<<PCT.test>>` was parsed, attached, and then dropped. Tests become a single
list of `TestDefinition`; `testCount` and `pctTestCount` become getters over it,
so they cannot drift. They are `@JsonIgnore` because a derived value written to
JSON cannot be read back, and these files round-trip through
`PCTReportProviderLoader`.

**A standard for writing the documentation**, in `docs/standards/`.

## What the documentation says

Examples and behaviour are taken from the PCT assertions in each file rather
than from the signature. Signatures say what the types are; the tests say what
actually happens, and that is usually the part worth writing down:

- `round` is half-even — `17.5` gives `18`, `16.5` gives `16`
- `dateDiff` counts calendar boundaries, not elapsed time: `2016-02-01` to
  `2016-02-29` is **0** months, while `2015-12-31T23:59:59` to
  `2016-01-01T00:00:01` is **1** year
- the `[0..1]` comparison overloads return `false` when an operand is absent
  rather than propagating emptiness, so a missing value never compares
- `indexOf` returns `-1`, not empty, so an `isEmpty` guard silently never fires
- `at` returns `[1]` and errors out of range, unlike `first`/`last` at `[0..1]`
- the date extractors error rather than return empty when a component is
  absent — which is what the `has*` family exists for
- `pow` returns floating point even for integer arguments
- `getIfAbsentPutWithKey` mutates the map

Cross-references use full signatures, since `substring` has two overloads and
`contains` exists in both the string and collection packages. They were taken
from the generated `FUNCTIONS_*.json`, not written by hand, and all 464 resolve.

## Two things reviewers should know

**Documentation shifts line numbers, and some tests assert them.** They come in
three unrelated shapes — error-message strings, Java integer assertions, and
self-referential `assertError` calls that assert the line they sit on — so no
single grep finds them all. `TestSourceNavigation` moves with `print.pure`; the
affected files are listed in the standard.

**The tests that exercise this do not run in `legend-pure-m3-core`.** Both
`Test_*_PCT` and `TestCoreFunctions` live in the runtime engine modules, and
`TestCoreFunctions` does not match `*_PCT`, so a filtered run skips it. A green
m3-core build says nothing about the position assertions above.

## Verification

- `mvn clean install -pl legend-pure-core/legend-pure-m3-core -am`
- `mvn test -pl legend-pure-runtime/legend-pure-runtime-java-engine-interpreted`
  (whole module, not filtered)
- 309/309 signatures documented, 464/464 cross-references resolve, 0 checkstyle
  violations

Documentation is parse-time sugar over `doc.doc`, so no PCT result changes.

## Follow-up, not in this PR

Neither renderer in `legend-engine` handles Markdown. `PCT_to_SimpleHTML`
HTML-escapes into a `<td>` with no `<pre>`, so newlines collapse.
`DocumentationHelper` passes documentation through `WordUtils.wrap`, which does
not reset its column counter at existing newlines — so in any doc longer than
80 characters it breaks lines mid-token, including inside fenced code blocks.
Both are small fixes on the engine side and neither affects this repo.

---

# Review round 2 — accuracy audit and PCT index changes

## What prompted it

`fold` was documented as reducing "to a single value". The signature is
`fold<T,V|m>(value:T[*], func:…, accumulator:V[m]):V[m]`, so the result carries the
accumulator's multiplicity and may be a collection — `testFoldCollectionAccumulator`
asserts exactly that. Rather than patch the one case, every documented file was
audited against its declared signature, its body, and its own test assertions.

**Corrections found (6 files):**

| File | Was wrong |
|---|---|
| `fold` | "reduces to a single value" — result is `V[m]` |
| `replaceAll` | Documented *merge* semantics; it discards the original map's entries |
| `reactivate` | Singular prose on an `Any[*]` return |
| `getUpperBound` | Implied a value is returned where the tests assert the call fails |
| `hasSubsecond` | Claimed a predicate "fails" on a missing component |
| `dateDiff` | "negative when earlier" — `WEEKS` is not antisymmetric |

`replaceAll` is the significant one: `testReplaceAll` seeds a map with keys 1 and 2,
supplies keys 1 and 3, and asserts the result has size **2**, not 3. Following the old
documentation would silently drop entries.

## Reviewer feedback

All 21 review threads addressed. Beyond the direct wording fixes:

- **`cast` / `toMultiplicity`** — reworded per review: a cast need not narrow (it may
  widen, or move to an unrelated type, since Pure supports multiple inheritance) and
  **never changes the value**, only the declared type.
- **`plus` / `minus` / `times` / `divide`** — the "missing documentation" on the
  Integer/Number variants had a single cause: those declarations still carried a legacy
  `doc.doc=` tagged value, and a `'''` literal alongside an explicit `doc.doc` is a
  compile error. The tagged value is removed and `PCT.grammarDoc` retained. Empty and
  single-element behaviour is now documented, verified against **both** engines: empty
  gives `0` / `0` / `1`, and `minus` on a single value **negates** (`minus(1)` is `-1`).
- **Inequality operators** — the `>` / `<` / `<=` documentation blocks sat on the
  `(Number[1], Number[1])` overload while describing operator syntax, other operand
  types, and empty-operand behaviour. Empty-operand prose moved to the `[0..1]`
  overloads, cross-type framing and non-numeric examples to the per-type overloads.
  `greaterThanEqual` was already anchored on `[0..1]` and keeps its empty-operand
  paragraph.

## Scope note — this round changes the PCT function index

Reviewer feedback on `elementToPath` exposed a structural problem: the documentation
sat on the `Function<Any>[1]` variant, kept only for backward compatibility, while the
`PackageableElement` variants are fundamental. Those variants carried no
`<<PCT.function>>` stereotype, so documentation moved there would never have reached
`FUNCTIONS_*.json`.

Resolved by indexing them rather than leaving the documentation unpublished:

- All seven `elementToPath` overloads now carry `<<PCT.function, PCT.platformOnly>>`,
  and the file is ordered `PackageableElement[1]` → `Type[1]` → `Function<Any>[1]`.
- `elementPath` and `lenientPathToElement` moved to their own files and were
  stereotyped. Both had to move: `FunctionsGeneration` permits only **one PCT function
  name per source file**. Their tests moved with them, since tests are attributed to a
  function by source id.
- `pathToElement(String[1], String[1])` stereotyped.

**PCT index: 312 → 322 signatures.** All ten additions are `PCT.platformOnly`, so none
enters the cross-engine adapter matrix and no adapter results change. The affected
surface is `FUNCTIONS_*.json` and the published documentation. Two new files under
`platform/pure/**` required bumping the count in
`TestClassLoaderCodeStorage#testGetUserFiles` (241 → 243).

## Verification

- 302 documentation blocks, 0 structural defects
- 591 cross-references, **0 unresolved and 0 outside the PCT index**
- No file mixes `platformOnly` with regular PCT functions, declares more than one PCT
  function name, or places a `<<PCT.test>>` in a platform-only file
- No changed file carries a self-referential `assertError` position assertion

## Engine defects found, deliberately not addressed here

Two genuine cross-engine divergences surfaced while verifying documentation claims.
Both are undocumented because there is no single true behaviour to state, and both
share a root cause — no PCT test covers the case:

- **`monthNumber`** — compiled `CoreHelper.monthNumber()` has no `hasMonth()` guard and
  returns `-1` (`Year.getMonth()`), while `dayOfMonth()` and `hour()` in the same class
  do guard and throw. Interpreted `MonthNumber` throws `Cannot get month for …`. So
  `%2015->monthNumber()` errors on one engine and returns `-1` on the other.
  `monthNumber.pure` is the only date extractor with no `assertError` test.
- **Division by zero** — compiled `CompiledSupport.divide()` throws
  `Cannot divide X by zero`; interpreted `Divide` has no guard and reaches
  `NumericAccumulator.divide`, giving `Infinity`/`NaN` or a raw `ArithmeticException`.

Both are now tracked separately: #1309 (`monthNumber`) and #1310 (division by zero).

### DIFF
--- a/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/collection/iteration/fold.pure
+++ b/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/collection/iteration/fold.pure
@@ -16,28 +16,33 @@ import meta::pure::test::pct::*;
 import meta::pure::functions::collection::tests::fold::*;
 
 '''
-Reduces a collection to a single value by combining its elements one at a time.
+Combines the elements of a collection into an accumulated result, one element at a time.
 
 `func` receives the current element and the accumulated value, and returns the new accumulated
 value. It is applied left to right, so the first element is combined with `accumulator` first. An
 empty collection returns `accumulator` unchanged.
+
+The accumulator is not restricted to a single value. Its multiplicity `m` is carried through to the
+result, so folding into a collection accumulator yields a collection — `fold` builds up a result of
+whatever shape the seed has, rather than always reducing to one value.
 
 Note the argument order: the element comes first, the accumulator second. This matters when the
 operation is not commutative — `{x, y | $y + $x}` concatenates in collection order, while
 `{x, y | $x + $y}` reverses it.
 
 **Parameters**
-- `value` — the collection to reduce.
+- `value` — the collection to fold over.
 - `func` — combines an element with the accumulated value.
 - `accumulator` — the starting value, and the result for an empty collection.
 
-**Returns** the final accumulated value.
+**Returns** the final accumulated value, with the same multiplicity as `accumulator`.
 
 **Examples**
 ```pure
-[1, 2, 3, 4]->fold({x, y | $x + $y}, 0)         // 10
-[1, 2, 3, 4]->fold({x, y | $x + $y}, 7)         // 17  -- the seed participates
-['a', 'b', 'c']->fold({x, y | $y + $x}, '')     // 'abc'
+[1, 2, 3, 4]->fold({x, y | $x + $y}, 0)              // 10
+[1, 2, 3, 4]->fold({x, y | $x + $y}, 7)              // 17  -- the seed participates
+['a', 'b', 'c']->fold({x, y | $y + $x}, '')          // 'abc'
+[1, 2, 3, 4]->fold({x, y | $y->add($x)}, [-1, 0])    // [-1, 0, 1, 2, 3, 4]  -- collection accumulator
 ```
 
 **See also** `map(T[m], Function<{T[1]->V[1]}>[1]):V[m]`,
--- a/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/collection/iteration/fold.pure
+++ b/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/collection/iteration/fold.pure
@@ -15,11 +15,37 @@
 import meta::pure::test::pct::*;
 import meta::pure::functions::collection::tests::fold::*;
 
+'''
+Reduces a collection to a single value by combining its elements one at a time.
+
+`func` receives the current element and the accumulated value, and returns the new accumulated
+value. It is applied left to right, so the first element is combined with `accumulator` first. An
+empty collection returns `accumulator` unchanged.
+
+Note the argument order: the element comes first, the accumulator second. This matters when the
+operation is not commutative — `{x, y | $y + $x}` concatenates in collection order, while
+`{x, y | $x + $y}` reverses it.
+
+**Parameters**
+- `value` — the collection to reduce.
+- `func` — combines an element with the accumulated value.
+- `accumulator` — the starting value, and the result for an empty collection.
+
+**Returns** the final accumulated value.
+
+**Examples**
+```pure
+[1, 2, 3, 4]->fold({x, y | $x + $y}, 0)         // 10
+[1, 2, 3, 4]->fold({x, y | $x + $y}, 7)         // 17  -- the seed participates
+['a', 'b', 'c']->fold({x, y | $y + $x}, '')     // 'abc'
+```
+
+**See also** `map(T[m], Function<{T[1]->V[1]}>[1]):V[m]`,
+`filter(T[*], Function<{T[1]->Boolean[1]}>[1]):T[*]`,
+`meta::pure::functions::string::joinStrings(String[*], String[1]):String[1]`
+'''
 native function
     <<PCT.function>>
-    {
-        doc.doc=''
-    }
     meta::pure::functions::collection::fold<T,V|m>(value:T[*], func:Function<{T[1],V[m]->V[m]}>[1], accumulator:V[m]):V[m];
 
 function <<PCT.test>> meta::pure::functions::collection::tests::fold::testIntegerSum<Z|y>(f:Function<{Function<{->Z[y]}>[1]->Z[y]}>[1]):Boolean[1]
--- a/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/date/operation/dateDiff.pure
+++ b/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/date/operation/dateDiff.pure
@@ -14,6 +14,40 @@
 
 import meta::pure::test::pct::*;
 
+'''
+The number of duration-unit boundaries between two dates, signed.
+
+**This counts calendar boundaries crossed, not elapsed time** — the single most common source of surprise
+here. `%2016-02-01` to `%2016-02-29` is `0` months, because both fall in the same calendar month, even
+though they are 28 days apart. Conversely `%2015-12-31T23:59:59` to `%2016-01-01T00:00:01` is `1` year,
+two seconds apart, because a year boundary lies between them.
+
+If you need elapsed duration rather than boundaries crossed, difference in a finer unit and convert.
+
+The result is **signed and directional**: `$d1->dateDiff($d2, …)` is positive when `d2` is later, negative
+when earlier, and `0` for two dates in the same interval.
+
+**Parameters**
+- `d1` — the starting date.
+- `d2` — the ending date.
+- `du` — the unit whose boundaries are counted.
+
+**Returns** the signed number of boundaries from `d1` to `d2`.
+
+**Examples**
+```pure
+%2015->dateDiff(%2016, DurationUnit.YEARS)                    // 1
+%2016->dateDiff(%2015, DurationUnit.YEARS)                    // -1
+%2015->dateDiff(%2015, DurationUnit.YEARS)                    // 0
+%2016-02-01->dateDiff(%2016-02-29, DurationUnit.MONTHS)       // 0   -- same month
+%2016-02-01->dateDiff(%2016-03-01, DurationUnit.MONTHS)       // 1
+%2015-01-01T00:00:00->dateDiff(%2015-12-31T23:59:59, DurationUnit.YEARS)   // 0
+%2015-12-31T23:59:59->dateDiff(%2016-01-01T00:00:01, DurationUnit.YEARS)   // 1
+```
+
+**See also** `adjust(Date[1], Integer[1], DurationUnit[1]):Date[1]`,
+`datePart(Date[1]):Date[1]`
+'''
 native function <<PCT.function>> meta::pure::functions::date::dateDiff(d1:Date[1], d2:Date[1], du:DurationUnit[1]):Integer[1];
 
 function <<PCT.test>> meta::pure::functions::date::tests::testDateDiffYears<Z|y>(f:Function<{Function<{->Z[y]}>[1]->Z[y]}>[1]):Boolean[1]
--- a/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/date/operation/dateDiff.pure
+++ b/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/date/operation/dateDiff.pure
@@ -24,8 +24,10 @@ two seconds apart, because a year boundary lies between them.
 
 If you need elapsed duration rather than boundaries crossed, difference in a finer unit and convert.
 
-The result is **signed and directional**: `$d1->dateDiff($d2, …)` is positive when `d2` is later, negative
-when earlier, and `0` for two dates in the same interval.
+The result is **signed and directional**: `$d1->dateDiff($d2, …)` is positive when `d2` is later, zero or
+negative when earlier, and `0` for two dates in the same interval. Swapping the two dates does not always
+just flip the sign: `%2015-07-04->dateDiff(%2015-07-05, DurationUnit.WEEKS)` is `1`, while
+`%2015-07-05->dateDiff(%2015-07-04, DurationUnit.WEEKS)` is `0`.
 
 **Parameters**
 - `d1` — the starting date.
--- a/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/date/operation/dateDiff.pure
+++ b/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/date/operation/dateDiff.pure
@@ -20,9 +20,9 @@ The number of duration-unit boundaries between two dates, signed.
 **This counts calendar boundaries crossed, not elapsed time** — the single most common source of surprise
 here. `%2016-02-01` to `%2016-02-29` is `0` months, because both fall in the same calendar month, even
 though they are 28 days apart. Conversely `%2015-12-31T23:59:59` to `%2016-01-01T00:00:01` is `1` year,
-two seconds apart, because a year boundary lies between them.
+even though they are just two seconds apart, because a year boundary lies between them.
 
-If you need elapsed duration rather than boundaries crossed, difference in a finer unit and convert.
+If you need elapsed duration rather than boundaries crossed, compute the difference in a finer unit and then convert.
 
 The result is **signed and directional**: `$d1->dateDiff($d2, …)` is positive when `d2` is later, zero or
 negative when earlier, and `0` for two dates in the same interval. Swapping the two dates does not always
--- a/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/lang/cast/cast.pure
+++ b/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/lang/cast/cast.pure
@@ -16,17 +16,19 @@ import meta::pure::test::pct::*;
 import meta::pure::functions::lang::tests::cast::*;
 
 '''
-Reinterprets a value as a more specific type.
+Reinterprets a value under a different declared type.
 
 The target type is given as a type literal, written `@Type`, rather than as a value. The cast is
 checked at runtime: if the value is not actually an instance of that type, execution fails rather
 than returning an empty result. Multiplicity is carried through unchanged, so casting a `[*]` gives
 a `[*]`.
 
-Casting narrows the *static* type so that subtype properties become reachable; it does not convert
-between unrelated types. For numeric conversion use
-`meta::pure::functions::math::toFloat(Number[1]):Float[1]` or
-`meta::pure::functions::math::toDecimal(Number[1]):Decimal[1]`.
+**A cast never changes the value — only its declared type.** Narrowing to a subtype is the common
+case, and is what makes subtype properties reachable, but a cast need not narrow: it may widen to a
+more general type, which is unusual but does happen, or move to a type bearing no direct relation to
+the declared one, which is useful because Pure supports multiple inheritance. Contrast the numeric
+conversions `meta::pure::functions::math::toFloat(Number[1]):Float[1]` and
+`meta::pure::functions::math::toDecimal(Number[1]):Decimal[1]`, which do change the value.
 
 **Parameters**
 - `source` — the value to reinterpret.
--- a/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/lang/cast/cast.pure
+++ b/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/lang/cast/cast.pure
@@ -15,6 +15,35 @@
 import meta::pure::test::pct::*;
 import meta::pure::functions::lang::tests::cast::*;
 
+'''
+Reinterprets a value as a more specific type.
+
+The target type is given as a type literal, written `@Type`, rather than as a value. The cast is
+checked at runtime: if the value is not actually an instance of that type, execution fails rather
+than returning an empty result. Multiplicity is carried through unchanged, so casting a `[*]` gives
+a `[*]`.
+
+Casting narrows the *static* type so that subtype properties become reachable; it does not convert
+between unrelated types. For numeric conversion use
+`meta::pure::functions::math::toFloat(Number[1]):Float[1]` or
+`meta::pure::functions::math::toDecimal(Number[1]):Decimal[1]`.
+
+**Parameters**
+- `source` — the value to reinterpret.
+- `object` — a type literal, `@Type`, naming the target type.
+
+**Returns** the same value, typed as `T`.
+
+**Examples**
+```pure
+$anyValue->cast(@Person).firstName    // reach a Person property
+1->cast(@P8)                          // narrow to a Primitive extension
+```
+
+**See also** `match(Any[*], Function<{Nil[n]->T[m]}>[1..*]):T[m]`,
+`toMultiplicity(T[*], Any[z]):T[z]`,
+`meta::pure::functions::meta::instanceOf(Any[1], Type[1]):Boolean[1]`
+'''
 native function <<PCT.function>> meta::pure::functions::lang::cast<T|m>(source:Any[m], object:T[1]):T[m];
 
 
--- a/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/lang/cast/toMultiplicity.pure
+++ b/legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/lang/cast/toMultiplicity.pure
@@ -15,17 +15,20 @@
 import meta::pure::test::pct::*;
 
 '''
-Narrows a collection's multiplicity, checking the size at runtime.
+Reinterprets a collection under a different declared multiplicity, checking the size at runtime.
 
 The target multiplicity is given as a multiplicity literal, written `@[1]`, `@[0..1]`, `@[1..*]` and
-so on. Where `cast(Any[m], T[1]):T[m]` narrows the *type* and leaves multiplicity alone, this narrows
-the *multiplicity* and leaves the type alone.
+so on. Where `cast(Any[m], T[1]):T[m]` changes the declared *type* and leaves multiplicity alone,
+this changes the declared *multiplicity* and leaves the type alone.
+
+**The values themselves never change — only the declared multiplicity.** Narrowing is the common
+case, but it need not be a narrowing: the target may equally be the same multiplicity or a wider one.
 
 The check is enforced: a collection whose size does not fit the target fails with
 `Cannot cast a collection of size N to multiplicity [BOUNDS]` rather than being truncated or padded.
 
 **Parameters**
-- `source` — the collection to narrow.
+- `source` — the collection to reinterpret.
 - `object` — a multiplicity literal, `@[…]`, naming the target multiplicity.
 
 **Returns** the same values, with multiplicity `z`.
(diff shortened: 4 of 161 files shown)


legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/collection/iteration/fold.pure
```diff
@@ -15,11 +15,37 @@
 import meta::pure::test::pct::*;
 import meta::pure::functions::collection::tests::fold::*;
 
+'''
+Reduces a collection to a single value by combining its elements one at a time.
```
kevin-m-knight-gs: The result need not be a single value - it could be a collection.

rafaelbey: Make sense - updating...

rafaelbey: Fixed in fe9d3af4f.

You are right — `accumulator:V[m]` carries `m` through to the `V[m]` return, so the result has whatever shape the seed has. The summary now reads "Combines the elements of a collection into an accumulated result, one element at a time", the **Returns** line states the multiplicity is the accumulator's, and there is a new paragraph making the point explicitly. I also added an example taken from `testFoldCollectionAccumulator`, which already asserted exactly this:

```pure
[1, 2, 3, 4]->fold({x, y | $y->add($x)}, [-1, 0])    // [-1, 0, 1, 2, 3, 4]  -- collection accumulator
```

Since this looked like it could be a systematic problem rather than a one-off, I audited all 151 documented files against their declared signatures, bodies and in-file test assertions. That turned up five more, the worst being `replaceAll`, which documented *merge* semantics while the function actually discards the original map's entries — `testReplaceAll` asserts the result has size 2 where a merge would give 3, so anyone following the doc would have silently lost data. The others were `reactivate` (singular prose on an `Any[*]` return), `getUpperBound` (implied a value is returned where the tests assert the call fails), `hasSubsecond` (claimed a predicate "fails" on a missing component) and `dateDiff` ("negative when earlier", but `WEEKS` is not antisymmetric — `%2015-07-05->dateDiff(%2015-07-04, DurationUnit.WEEKS)` is `0`).

Verified locally: m3-core 1684 passed, interpreted engine 1179 passed, zero failures.

Separately, the audit surfaced something that is **not** a documentation issue and is not addressed here: `monthNumber` diverges between the engines. Compiled `CoreHelper.monthNumber()` is `return date.getMonth()` with no `hasMonth()` guard — while `dayOfMonth()` and `hour()` in the same file do guard and throw — and `Year.getMonth()` returns `-1`. The interpreted `MonthNumber` throws `Cannot get month for ...`. So `%2015->monthNumber()` errors on interpreted and returns `-1` on compiled. It appears to have gone unnoticed because `monthNumber.pure` is the only date extractor without an `assertError` test. Happy to raise it separately.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/date/operation/dateDiff.pure
```diff
@@ -14,6 +14,42 @@
 
 import meta::pure::test::pct::*;
 
+'''
+The number of duration-unit boundaries between two dates, signed.
+
+**This counts calendar boundaries crossed, not elapsed time** — the single most common source of surprise
+here. `%2016-02-01` to `%2016-02-29` is `0` months, because both fall in the same calendar month, even
+though they are 28 days apart. Conversely `%2015-12-31T23:59:59` to `%2016-01-01T00:00:01` is `1` year,
+two seconds apart, because a year boundary lies between them.
```
kevin-m-knight-gs:
```suggestion
though they are 28 days apart. Conversely `%2015-12-31T23:59:59` to `%2016-01-01T00:00:01` is `1` year,
even though they are just two seconds apart, because a year boundary lies between them.
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/date/operation/dateDiff.pure
```diff
@@ -14,6 +14,42 @@
...
+
+If you need elapsed duration rather than boundaries crossed, difference in a finer unit and convert.
```
kevin-m-knight-gs:
```suggestion
If you need elapsed duration rather than boundaries crossed, compute the difference in a finer unit and then convert.
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/lang/cast/cast.pure
```diff
@@ -15,6 +15,35 @@
...
+
+The target type is given as a type literal, written `@Type`, rather than as a value. The cast is
+checked at runtime: if the value is not actually an instance of that type, execution fails rather
+than returning an empty result. Multiplicity is carried through unchanged, so casting a `[*]` gives
+a `[*]`.
+
+Casting narrows the *static* type so that subtype properties become reachable; it does not convert
+between unrelated types. For numeric conversion use
+`meta::pure::functions::math::toFloat(Number[1]):Float[1]` or
+`meta::pure::functions::math::toDecimal(Number[1]):Decimal[1]`.
```
kevin-m-knight-gs: Casting does not have to be done to a more specific type, though that is probably the most common case. It can be done to a more general type (which is unusual, but does happen occasionally) or to a type that bears no direct relation to the declared type (which can be useful since Pure supports multiple inheritance).

What it _never_ does is change the value itself. Casting only ever changes the declared type, unlike numeric conversions which can change the value.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/lang/cast/toMultiplicity.pure
```diff
@@ -14,6 +14,31 @@
 
 import meta::pure::test::pct::*;
 
+'''
+Narrows a collection's multiplicity, checking the size at runtime.
+
+The target multiplicity is given as a multiplicity literal, written `@[1]`, `@[0..1]`, `@[1..*]` and
+so on. Where `cast(Any[m], T[1]):T[m]` narrows the *type* and leaves multiplicity alone, this narrows
+the *multiplicity* and leaves the type alone.
```
kevin-m-knight-gs: As with cast, while narrowing is the most common case, it doesn't have to be a narrowing.

Also, it's probably worth emphasizing that the value itself never changes, only the declared multiplicity.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/meta/graph/elementToPath.pure
```diff
@@ -15,6 +15,31 @@
 import meta::pure::functions::meta::tests::model::*;
 import meta::pure::test::pct::*;
 
+'''
+The fully-qualified path of a model element, as a string.
```
kevin-m-knight-gs: I think the documentation is attached to the wrong variant of elementToPath. The fundamental variants are the native ones that take a PackageableElement as the argument. I think those should hold the main docs.

The variants for Function and Type are here principally for backward compatibility, since they used to be (but are no longer) subtypes of PackageableElement.

Also, there is a distinct but related function, elementPath, in this file which may warrant its own doc.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/meta/graph/pathToElement.pure
```diff
@@ -15,6 +15,30 @@
 import meta::pure::functions::meta::tests::model::*;
 import meta::pure::test::pct::*;
 
+'''
+Looks up a model element by its fully-qualified path.
+
+The inverse of `elementToPath(Function<Any>[1]):String[1]`, and the way to reach an element named at
```
kevin-m-knight-gs:
```suggestion
The inverse of `elementToPath(PackageableElement[1]):String[1]`, and the way to reach an element named at
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/meta/graph/pathToElement.pure
```diff
@@ -15,6 +15,30 @@
...
+runtime — from configuration, user input, or generated code — rather than written as a literal.
+
+The path is `::`-separated. **A path that resolves to nothing is an error, not an empty result**, since
+the return multiplicity is `[1]`; use `lenientPathToElement` in the same package when a missing element
+should yield empty instead.
```
kevin-m-knight-gs: I think it may be worth documenting both lenientPathToElement and the variant of pathToElement that takes a separator as an argument.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/meta/profile/tag.pure
```diff
@@ -14,4 +14,30 @@
 
 import meta::pure::test::pct::*;
 
+'''
+Looks up a tag on a profile by name.
+
+Tags are the `{Profile.name='value'}` annotations attached to model elements; where a stereotype is a
+bare marker, a tag carries a string value. This resolves the name to the `Tag` object, which is what you
+match a `TaggedValue`'s `tag` against — comparing names alone would ignore which profile the tag belongs
+to, and two profiles may define tags of the same name.
```
kevin-m-knight-gs:
```suggestion
A TaggedValue annotation on a model element, `{Profile.name='value'}`, associates a Tag with a string value (unlike a stereotype, which is a bare marker). This resolves the name to the `Tag` object, which is what you
match a `TaggedValue`'s `tag` against — comparing names alone would ignore which profile the tag belongs
to, and two profiles may define tags of the same name.
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/math/operation/times.pure
```diff
@@ -23,10 +23,28 @@ native function
     }
     meta::pure::functions::math::times(ints:Integer[*]):Integer[1];
 
+'''
```
kevin-m-knight-gs, line 52: The doc is missing for the Integer variant of times. Also, it may be worthwhile explaining the result when the collection is empty or has just one value.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/math/operation/plus.pure
```diff
@@ -23,10 +23,28 @@ native function
     }
     meta::pure::functions::math::plus(ints:Integer[*]):Integer[1];
 
+'''
```
kevin-m-knight-gs, line 52: The doc is missing for the Integer version of plus. Also, it may be worthwhile explaining the result when the collection is empty or has just one value.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/math/operation/minus.pure
```diff
@@ -23,10 +23,28 @@ native function
     }
     meta::pure::functions::math::minus(ints:Integer[*]):Integer[1];
 
+'''
```
kevin-m-knight-gs, line 53: The doc is missing for the Integer variant. Also, it may be worthwhile explaining the result when the collection is empty or has just one value.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/math/operation/divide.pure
```diff
@@ -23,6 +23,20 @@ native function
     }
     meta::pure::functions::math::divide(left:Number[1], right:Number[1]):Float[1];
 
+'''
```
kevin-m-knight-gs, line 52: The doc is missing for the Number variant.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/meta/type/enum/enumName.pure
```diff
@@ -15,6 +15,26 @@
 import meta::pure::functions::meta::tests::model::*;
 import meta::pure::test::pct::*;
 
+'''
+The simple name of an enumeration, without its package.
+
+Takes the enumeration itself, not one of its values, so it answers "what is this enum called" rather
```
kevin-m-knight-gs:
```suggestion
Takes the enumeration itself, not one of its values, so it answers "what is this enumeration called" rather
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/meta/type/enum/enumName.pure
```diff
@@ -15,6 +15,26 @@
...
+than "which value is this". The result is the bare name — use
+`elementToPath(Function<Any>[1]):String[1]` for the fully-qualified path.
```
kevin-m-knight-gs:
```suggestion
`elementToPath(PackageableElement[1]):String[1]` for the fully-qualified path.
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/meta/type/enum/enumName.pure
```diff
@@ -15,6 +15,26 @@
...
+
+**Returns** its simple name.
+
+**Examples**
+```pure
+CC_GeographicEntityType->enumName()   // 'CC_GeographicEntityType'
+```
+
+**See also** `enumValues(Enumeration<T>[1]):T[*]`,
+`elementToPath(Function<Any>[1]):String[1]`
```
kevin-m-knight-gs:
```suggestion
`elementToPath(PackageableElement[1]):String[1]`
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/meta/type/enum/enumValues.pure
```diff
@@ -15,6 +15,27 @@
 import meta::pure::functions::meta::tests::model::*;
 import meta::pure::test::pct::*;
 
+'''
+Every value of an enumeration, in declaration order.
+
+The way to enumerate an enum's members without naming them — for building a lookup, validating input,
+or iterating all cases. The enumeration may be held in a variable, so the enum need not be known when
+the code is written.
```
kevin-m-knight-gs:
```suggestion
The way to enumerate an enumeration's members without naming them — for building a lookup, validating input,
or iterating all cases. The enumeration may be held in a variable, so it need not be known when
the code is written.
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/meta/type/generalizations.pure
```diff
@@ -15,6 +15,31 @@
 import meta::pure::functions::meta::tests::generalizations::*;
 import meta::pure::test::pct::*;
 
+'''
+A type together with all of its supertypes, most specific first.
```
kevin-m-knight-gs:
```suggestion
A type together with all of its supertypes, in inheritance order. That is, the order in which types are searched when trying to resolve something (such as a property) through inheritance.
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/essential/string/toString/toString.pure
```diff
@@ -16,11 +16,32 @@ import meta::pure::test::pct::*;
 import meta::pure::functions::string::tests::toString::*;
 
 // Used by List
+'''
+The plain textual form of any value.
+
+This is the human-readable rendering: strings come back as their own contents, with no surrounding
+quotes. Use `toRepresentation(Any[1]):String[1]` when you want the form a value would take in Pure
+source, with strings quoted and escaped.
+
```
kevin-m-knight-gs, line 29:
```suggestion
This is the human-readable rendering: strings come back as their own contents, with no surrounding
quotes or escapes. This can yield ambiguous results; e.g., `1->toString()` and `'1'->toString() both
give the same result.  Use `toRepresentation(Any[1]):String[1]` when you want the form a value would
take in Pure source, with strings quoted and escaped.

If a Class defines a qualified property toString():String[1], it will be called by this function to generate
string renderings for its instances.
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/boolean/inequality/greaterThan.pure
```diff
@@ -14,10 +14,37 @@
...
+'''
+Whether the first operand succeeds the second, according to the type's natural order.
+
+Written `>` in Pure: `$a > $b`. Overloads cover `Number`, `Date`, `String` and `Boolean`, so the
+same operator orders numerically, chronologically, lexicographically, or with `false` before `true`.
+
+An absent operand makes the result `false`, **not** empty — the comparison does not propagate
+emptiness the way arithmetic does. So a missing value never satisfies the comparison, and
+`$a > $b` is false whenever either side is empty. Test with `isEmpty()` first when you need to
+distinguish "absent" from "present but not greater than".
```
kevin-m-knight-gs: This paragraph looks like it belongs to the variant with [0..1] multiplicity for the parameters.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/boolean/inequality/greaterThan.pure
```diff
@@ -14,10 +14,37 @@
 
 import meta::pure::test::pct::*;
 
+'''
+Whether the first operand succeeds the second, according to the type's natural order.
+
+Written `>` in Pure: `$a > $b`. Overloads cover `Number`, `Date`, `String` and `Boolean`, so the
+same operator orders numerically, chronologically, lexicographically, or with `false` before `true`.
```
kevin-m-knight-gs: This seems to be conflating this function, which is greaterThan for Number, with '>' syntax and greaterThan for other types.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/boolean/inequality/greaterThan.pure
```diff
@@ -14,10 +14,37 @@
...
+- `right` — the second operand.
+
+**Returns** `true` when `left` is greater than `right`.
+
+**Examples**
+```pure
+1 > 2
+'a' > 'b'
+%2015-01-01 > %2016-01-01
+false > true
```
kevin-m-knight-gs: I don't think these examples belong in the documentation for greaterThan(Number[1],Number[1]):Boolean[1].

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/boolean/inequality/greaterThan.pure
```diff
@@ -27,76 +52,187 @@ function
...
+
+**See also** `greaterThan(Number[1], Number[1]):Boolean[1]`
+'''
 function <<PCT.function>> meta::pure::functions::boolean::greaterThan(left:String[1], right:String[1]):Boolean[1]
 {
     lessThan($right, $left)
 }
 
+'''
+Compares two dates chronologically — true when `left` falls strictly after `right`.
```
kevin-m-knight-gs:
```suggestion
Compares two dates chronologically. For dates of the same granularity, this is true when `left` falls strictly after `right`. The general case is that it is true when `left` starts strictly after the start of `right`, or when they start at the same time but `left` ends strictly before `right`.
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/boolean/inequality/greaterThanEqual.pure
```diff
@@ -14,10 +14,42 @@
 
 import meta::pure::test::pct::*;
 
+'''
```
kevin-m-knight-gs, line 17: I tend to think the strict [1] multiplicity versions should get the main docs, while the others should be documented in terms of the strict ones.

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/boolean/inequality/greaterThanEqual.pure
```diff
@@ -26,76 +58,183 @@ function
...
+
+**See also** `greaterThanEqual(Number[1], Number[1]):Boolean[1]`
+'''
 function <<PCT.function>> meta::pure::functions::boolean::greaterThanEqual(left:String[1], right:String[1]):Boolean[1]
 {
     lessThanEqual($right, $left)
 }
 
+'''
+Compares two dates chronologically — true when `left` falls at or after `right`.
```
kevin-m-knight-gs:
```suggestion
Compares two dates chronologically. For dates of the same granularity, this is true when `left` falls at or after `right`. The general case is that it is true when `left` starts strictly after the start of `right`, or when they start at the same time and `left` ends before or at the same time as `right`.
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/boolean/inequality/lessThan.pure
```diff
@@ -14,86 +14,222 @@
...
+The first operand may be absent; absent operands are treated as described in
+`lessThan(Number[0..1], Number[0..1]):Boolean[1]`.
+'''
 function <<PCT.function>> meta::pure::functions::boolean::lessThan(left:Boolean[0..1], right:Boolean[1]):Boolean[1]
 {
     $left->isNotEmpty() && lessThan($left->toOne(), $right)
 }
 
+'''
+Compares two dates chronologically — true when `left` falls strictly before `right`.
```
kevin-m-knight-gs:
```suggestion
Compares two dates chronologically. For dates of the same granularity, this is true when `left` falls strictly before `right`. The general case is that it is true when `left` starts strictly before the start of `right`, or when they start at the same time and `left` ends strictly after the end of `right`.
```

legend-pure-core/legend-pure-m3-core/src/main/resources/platform/pure/grammar/functions/boolean/inequality/lessThanEqual.pure
```diff
@@ -14,86 +14,223 @@
...
+The first operand may be absent; absent operands are treated as described in
+`lessThanEqual(Number[0..1], Number[0..1]):Boolean[1]`.
+'''
 function <<PCT.function>> meta::pure::functions::boolean::lessThanEqual(left:Boolean[0..1], right:Boolean[1]):Boolean[1]
 {
     $left->isNotEmpty() && lessThanEqual($left->toOne(), $right)
 }
 
+'''
+Compares two dates chronologically — true when `left` falls at or before `right`.
```
kevin-m-knight-gs:
```suggestion
Compares two dates chronologically. For dates of the same granularity, this is true when `left` falls at or before `right`. The general case is that it is true when `left` starts strictly before the start of `right`, or when they start at the same time and `left` ends at the same time as or after the end of `right`.
```

kevin-m-knight-gs (approved):

PR #1322, opened 2026-08-16
Default RuntimeOptions consistently and snapshot them once per VM

rafaelbey:
## Problem

`RuntimeOptions` had two different "no options were supplied" defaults:

| Path | Default |
|---|---|
| `PureRuntimeBuilder` | `systemPropertyOptions("pure.options.")` |
| `CompiledExecutionSupport` (`options == null`) | `name -> false` |

Three of `CompiledExecutionSupport`'s convenience constructors pass `null`, so every standalone compiled execution silently saw *no* options — including the bootstrap code emitted by `ExternalClassBuilder`, `PureTestBuilderCompiled`, and downstream callers using the shorter overloads. A `pure.options.*` property was honoured by a builder-built runtime and ignored by a standalone one.

Separately, `systemPropertyOptions(prefix)` is `name -> Boolean.getBoolean(prefix + name)`: a string concat plus a synchronized `Properties` lookup on *every* `isOptionSet` call.

## Change

**One default everywhere.** `RuntimeOptions.defaultOptions()` resolves `pure.options.*` from a snapshot taken once per VM; `isOptionSet` becomes a set lookup. `PureRuntimeBuilder`, `PureRuntimeBuilder.withOptions(null)`, the `PureRuntime` constructor, and `CompiledExecutionSupport`'s null-fallback all route to it. The `ExternalClassBuilder`-generated bootstrap and `PureTestBuilderCompiled` inherit it through the existing overloads with no edits.

**Additive API.** `noOptionsSet()`, `systemPropertyOptions()` and `systemPropertyOptions(String)` are unchanged, so nothing downstream shifts behaviour on recompile. `systemPropertyOptions` stays live for callers that need per-call reads. New: `DEFAULT_OPTION_PREFIX`, `defaultOptions()`, `MutableRuntimeOptions`, and the package-private `CachedSystemPropertyOptions`.

**Timing contract.** The snapshot freezes at first use of the default options — in practice the first `new PureRuntimeBuilder(...)`. Properties set after that are not visible. This is documented on `defaultOptions()`.

**LSP no longer writes system properties.** `legend/setOption` previously did `System.setProperty` / `System.clearProperty` — a language-server RPC mutating global JVM state, and incompatible with a snapshot. The session now owns a `MutableRuntimeOptions`, seeded from `pure.options.*` once and mutated in memory. It is created eagerly and never replaced, so toggles survive a runtime rebuild. The trade, documented in the handler javadoc: the toggle is scoped to the session, so other code in the JVM reading `pure.options.*` directly will not observe it.

## Reviewer notes

- **`AbstractPureTestWithCoreCompiled.getOptions()` moves from `noOptionsSet()` to `defaultOptions()`.** Tests now resolve options like every other execution path. This means a `pure.options.*` in the surefire `argLine` reaches the ~30 test suites that call `getOptions()` — by design rather than by omission. No pom or surefire config sets one today (verified by grep), so current behaviour is unchanged.
- **`isOptionSet` has no consumer in this repo.** The native `meta::core::runtime::isOptionSet` was removed in `23fc77a63` along with the `legend-pure-functions` module; `PureRuntime.getOptions()` and `CompiledExecutionSupport.getRuntimeOptions()` are read by `legend-engine`. So this is a public-contract change, not a hot-path optimisation with an in-repo benchmark.

## Testing

`CachedSystemPropertyOptions` takes a `Properties` rather than reading `System.getProperties()` internally, so tests assert content against an instance they construct (no mocking framework, per ADR-003 — a real object). Consequences:

- No test mutates system properties, so the VM-wide snapshot is empty in every fork regardless of class order.
- Tests involving `defaultOptions()` assert **identity only** (`assertSame`), never contents, so none can depend on which test loaded the class first.

`TestRuntimeOptions` (8) covers prefix stripping, case-insensitive `true` matching `Boolean.getBoolean`, non-boolean and `false` values, the post-snapshot timing contract, and that the builder / `withOptions(null)` / explicit options resolve as expected. `TestMutableRuntimeOptions` (7) covers seeding, toggling, idempotence, the non-live `getSetOptions()` view, and that toggling never writes a property.

`LspEndToEndTest#setOption_rpc_togglesRuntimeOptionLive` now asserts `session.getPureRuntime().getOptions().isOptionSet(name)` — which passes only if the session → builder → runtime wiring is live — and that the system property stays null.

Full `mvn clean install` from the root: **BUILD SUCCESS**, 107 modules, **6803 tests, 0 failures, 0 errors**, 0 Checkstyle violations. That matters here because the `CompiledExecutionSupport` fallback flip touches every compiled-engine and PCT execution path.

### DIFF
--- a/legend-pure-core/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/runtime/CachedSystemPropertyOptions.java
+++ b/legend-pure-core/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/runtime/CachedSystemPropertyOptions.java
@@ -1,4 +1,4 @@
-// Copyright 2020 Goldman Sachs
+// Copyright 2026 Goldman Sachs
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/legend-pure-core/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/runtime/CachedSystemPropertyOptions.java
+++ b/legend-pure-core/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/runtime/CachedSystemPropertyOptions.java
@@ -1,0 +1,66 @@
+// Copyright 2020 Goldman Sachs
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package org.finos.legend.pure.m3.serialization.runtime;
+
+import org.eclipse.collections.api.factory.Sets;
+import org.eclipse.collections.api.set.MutableSet;
+import org.eclipse.collections.api.set.SetIterable;
+
+import java.util.Properties;
+
+/**
+ * {@link RuntimeOptions} backed by a snapshot of properties with a given prefix, taken once at construction.
+ * Properties set after the snapshot are not visible.
+ */
+class CachedSystemPropertyOptions implements RuntimeOptions
+{
+    static final RuntimeOptions DEFAULT = new CachedSystemPropertyOptions(System.getProperties(), RuntimeOptions.DEFAULT_OPTION_PREFIX);
+
+    private final SetIterable<String> setOptions;
+
+    CachedSystemPropertyOptions(Properties properties, String prefix)
+    {
+        this.setOptions = scanProperties(properties, prefix).toImmutable();
+    }
+
+    @Override
+    public boolean isOptionSet(String name)
+    {
+        return this.setOptions.contains(name);
+    }
+
+    @Override
+    public String toString()
+    {
+        return "<CachedSystemPropertyOptions " + this.setOptions + ">";
+    }
+
+    /**
+     * Names (with the prefix stripped) of the properties starting with the given prefix whose value is
+     * {@code true}, matching the case insensitive semantics of {@link Boolean#getBoolean}.
+     */
+    static MutableSet<String> scanProperties(Properties properties, String prefix)
+    {
+        MutableSet<String> result = Sets.mutable.empty();
+        properties.stringPropertyNames().forEach(key ->
+        {
+            if (key.startsWith(prefix) && Boolean.parseBoolean(properties.getProperty(key)))
+            {
+                result.add(key.substring(prefix.length()));
+            }
+        });
+        return result;
+    }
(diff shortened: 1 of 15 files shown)


legend-pure-core/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/runtime/CachedSystemPropertyOptions.java
```diff
@@ -0,0 +1,66 @@
+// Copyright 2020 Goldman Sachs
```
kevin-m-knight-gs:
```suggestion
// Copyright 2026 Goldman Sachs
```

legend-pure-core/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/runtime/MutableRuntimeOptions.java
```diff
@@ -0,0 +1,103 @@
+// Copyright 2020 Goldman Sachs
```
kevin-m-knight-gs:
```suggestion
// Copyright 2026 Goldman Sachs
```

legend-pure-core/legend-pure-m3-core/src/main/java/org/finos/legend/pure/m3/serialization/runtime/MutableRuntimeOptions.java
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * {@link RuntimeOptions} held in memory and togglable at runtime. It may be seeded from system properties,
+ * but it never writes them: toggling an option here is visible only to code holding this instance.
+ * <p>
+ * Safe for concurrent use.
+ */
+public final class MutableRuntimeOptions implements RuntimeOptions
+{
+    private final ConcurrentMutableMap<String, Boolean> options = ConcurrentHashMap.newMap();
```
kevin-m-knight-gs: You could potentially use ConcurrentHashSet here, since this is essentially being treated like a set.

legend-pure-core/legend-pure-m3-core/src/test/java/org/finos/legend/pure/m3/serialization/runtime/TestMutableRuntimeOptions.java
```diff
@@ -0,0 +1,120 @@
+// Copyright 2020 Goldman Sachs
```
kevin-m-knight-gs:
```suggestion
// Copyright 2026 Goldman Sachs
```

legend-pure-core/legend-pure-m3-core/src/test/java/org/finos/legend/pure/m3/serialization/runtime/TestRuntimeOptions.java
```diff
@@ -0,0 +1,115 @@
+// Copyright 2020 Goldman Sachs
```
kevin-m-knight-gs:
```suggestion
// Copyright 2026 Goldman Sachs
```

kevin-m-knight-gs (approved):

